Issue #1, opened 2018-08-08
Automate release to Firefox and Chrome web store


nicksnyder:
Right now when we want to release the browser extensions, a person needs to manually build and upload the extension to the respective stores. This isn't a terrible burden, but it would be ideal to automate this process.

Solution requirements:
- Releasing is an explicit step so we can control when it happens.
- Releasing is only a single step/command to indicate "I want to release"
- Releases are recorded as git tags in the repo so it is easy to know what code corresponds to a given release

Other context:
- The person performing the release shouldn't need to choose the version, it should be chosen automatically.
- The version number itself is mostly irrelevant (i.e. it is not important to be able to differentiate between major/minor/patch releases).
- Code in master should always be safe to release. We don't currently have enough automated tests to verify this assumption, so we are not (yet) comfortable auto-deploying on every commit to master.

There are multiple paths forward (tag to release, push to release branch, merge to release branch, bump version file, etc.), but I am going to defer discussion of those solutions until we have someone to own working on this.

Related:
- https://github.com/sourcegraph/browser-extension/pull/266
- https://github.com/sourcegraph/browser-extension/pull/311
- https://github.com/sourcegraph/sourcegraph/issues/12356

PR #11, opened 2018-08-09
ci: setup semantic-release

felixfbecker:
Use semantic-release to automatically release to Chrome and Firefox webstores whenever the `release` branch is being pushed to.

fixes https://github.com/sourcegraph/browser-extensions/issues/1
fixes https://github.com/sourcegraph/sourcegraph/issues/12356


README.md
```diff
@@ -132,10 +132,10 @@ Releases are also uploaded to the [GitHub releases page](https://github.com/sour
 
 Make sure that commit messages follow the [Conventional Commits Standard](https://conventionalcommits.org/) as the commit message will be used for the (public) release notes and to automatically determine the version number.
 
-To release the latest commit on master, ensure your master is up-to-date and run
+To release the latest commit on origin/master
 
 ```sh
-git push origin master:release
+git fetch && git push origin origin/master:refs/heads/release
```
felixfbecker: I considered this but decided against this - we want people to test locally, and release only tested commits. The commit that is tested is local `master` - if during testing, someone also merged a PR to master, this will silently include that PR in the release even though it was not tested by the person triggering the release. So even if a PR was just merged, I would prefer to release the (slightly outdated) local master.

nicksnyder: (I agree with what you said, just trying to think if there is a way to address both concerns)

felixfbecker: You could push `master` first (which would fail if it's behind). But honestly I don't think it's worth it, usually there are enough indicators if you have a new commit on master locally (terminal integrations, VS Code, etc). Our `git push origin master:docker-images/foo` system works the same way and we don't seem to be afraid of that scenario there (not even warned against in the docs). From my experience it has never happened/caused a problem

nicksnyder: The concern isn't about `master` being behind, it is about it being ahead. The solution would be to prevent direct commits to master and then git push would fail. Probably not worth it at this point so my commit can be reverted.

PR #16, opened 2018-08-11
feat: support CXP extensions

chrismwendt:
🎉 **CXP extensions** 🎉

This adds support for CXP extensions and brings the browser extension to near feature parity with the Sourcegraph UI.

---

**Feature flag in the popup:**

image

---

**Extensions page (shares UI components with https://sourcegraph.com/extensions):**

*TODO: don't show the option to add the extension to user/org/global settings*

image

---

**Line BG colors, annotations, and multiple CXP extensions:**

---

**Setting a GitHub token for private repositories:**

*TODO: provide more visual feedback that the token was saved and is valid*

Testing plan:

- `git checkout cxp-beta; git pull; npm run dev`
- Add/remove different extensions
- Switch between different Primary Sourcegraph URLs
- Enable/disable extensions
- Make sure Codecov shows up on https://github.com/ChacaS0/tempest/blob/master/cmd/purge.go
- Toggle the feature flag off and make sure existing functionality still works

**Known issues**

- CXP does not work on multiple files at a time (e.g. diff pages)

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle

cc @sqs


sqs:
Re: needing a GitHub token for private repos, that would be a big barrier to using the extension for users. Not insurmountable, but big. It would also add complexity since we'd want to use data from the Sourcegraph instance if available (instead of fetching from the GitHub API). Is it *absolutely* necessary? Can we obtain the file contents from the DOM or in some other way instead?

chrismwendt:
`document.querySelector('tbody').innerText` works for the current file (I'm surprised!)

What about other files (e.g. `.codecov.yml`)?

Embedding an iframe doesn't work (this error pops up: `Refused to frame 'https://github.com/gorilla/mux/blob/master/mux.go' because it violates the following Content Security Policy directive: "frame-src render.githubusercontent.com".`)

Cookies can't be used to hit `api.github.com` (it only accepts access tokens), so injecting a `<script>` into the DOM in order to get access to cookies doesn't help.

An OAuth app (hosted on Sourcegraph.com) might be possible and only require a few clicks from the user, but probably won't work on GitHub Enterprise.

GSuite settings could work for users that have it set up.

I tried to get the authenticated zip URL for the repository by injecting a `<script>` into the DOM but it's hidden behind a redirect, and I wasn't able to actually get the final URL programmatically (it shows up in the console, though).

I also tried the `webRequest` API, but was still unable to obtain the redirect URL. @KattMingMing could you try this to see if you're able to get it to work?

Other extensions require access tokens (e.g. Octotree), which makes me think there is probably no other way.

Any other ideas come to mind? 💭

chrome/extension/cxp.tsx
```diff
@@ -0,0 +1,95 @@
+// We want to polyfill first.
+// prettier-ignore
```
felixfbecker: what does prettier do to this?

extension/storage.ts
```diff
@@ -31,6 +34,51 @@ const getItem = (area: chrome.storage.StorageArea) => (
     callback: (items: StorageItems) => void
 ) => area.get(key, callback)
 
+const onChanged = (listener: (changes: Partial<StorageChange>, areaName: string) => void) => {
+    if (chrome && chrome.storage) {
```
felixfbecker: Should this be `self.chrome`/`window.chrome`? Referencing an undeclared variable is a TypeError in strict mode

extension/storage.ts
```diff
@@ -15,9 +16,11 @@ export interface Storage {
     getSync: (callback: (items: StorageItems) => void) => void
     getSyncItem: (key: keyof StorageItems, callback: (items: StorageItems) => void) => void
     setSync: (items: Partial<StorageItems>, callback?: (() => void) | undefined) => void
+    observeSync: <T extends keyof StorageItems>(key: T) => ReplaySubject<StorageItems[T]>
```
felixfbecker: If this returns a `ReplaySubject`, does that mean I can call `next()` on it to change values?

extension/storage.ts
```diff
@@ -31,6 +34,51 @@ const getItem = (area: chrome.storage.StorageArea) => (
...
+                listener(c, areaName)
+            }
+        }
+
+        page.addEventListener('message', handleChanges, false)
+    }
+}
+
+const observe = (area: chrome.storage.StorageArea) => <T extends keyof StorageItems>(key: T) => {
+    const itemsSubject = new ReplaySubject<StorageItems[T]>(1)
```
felixfbecker: Same as above, try to avoid Subjects

extension/types.ts
```diff
@@ -69,8 +93,9 @@ export const defaultStorageItems: StorageItems = {
     openFileOnSourcegraph: true,
     sourcegraphAnonymousUid: '',
     disableExtension: false,
-    useCXP: false,
-    featureFlags: featureFlagDefaults,
+    featureFlags: mapValues(featureFlags, v => v.default) as Record<Feature, boolean>,
```
felixfbecker: Why do you need to cast here?

package.json
```diff
@@ -67,19 +71,22 @@
...
+    "@sourcegraph/icons": "^1.15.0",
+    "@sqs/jsonc-parser": "^1.0.3",
     "@types/filesystem": "0.0.28",
-    "@types/react-dom": "16.0.7",
     "@types/uuid": "3.4.3",
     "bluebird": "^3.3.4",
     "bootstrap": "^4.0.0",
+    "cxp": "^6.4.0",
     "cypress-browser-extension-plugin": "^0.1.0",
+    "deepmerge": "^2.1.1",
```
felixfbecker: Did you try lodash equivalents?

package.json
```diff
@@ -67,19 +71,22 @@
...
     "tslint-webpack-plugin": "^1.0.0",
-    "typescript": "^2.9.2",
+    "typescript": "^3.0.1",
     "web-ext": "^2.5.0",
     "webpack": "^4.6.0"
   },
   "dependencies": {
     "@sourcegraph/codeintellify": "^3.6.4",
-    "@sourcegraph/react-loading-spinner": "0.0.6",
+    "@sourcegraph/extensions-client-common": "0.0.4-development",
```
felixfbecker: @sqs why does this have a `-development` prerelease tag? Seems unintentional?

tsconfig.json
```diff
@@ -1,5 +1,11 @@
 {
   "compilerOptions": {
+    "baseUrl": "",
+    "paths": {
+      "rxjs": ["node_modules/rxjs"],
+      "rxjs/*": ["node_modules/rxjs/*"]
+    },
```
felixfbecker: Why is this needed?

sqs:
Even having the current file be accessible without an access token is a huge improvement and enables a ton of functionality. That means that the browser extension can support didOpen 100%, although it can’t support xfiles.

We can use the tree-list API to support things like “tell me or activate me if this repo contains a file named ‘whatever’”. The tree-list API is what the “t” file finder uses. It is undocumented and is on GitHub.com not api.github.com.

My only other idea is to use the raw link and obtain the contents from that. Might be blocked (like the archive link) though.

felixfbecker:
What extensions need `xfiles`?

app/backend/extensions.ts
```diff
@@ -0,0 +1,202 @@
...
+                    queryGraphQL(getContext({ repoKey: '', isRepoSpecific: false }), request, variables, [url])
+                )
+            ),
+        icons: {
+            Loader: Loader as React.ComponentType<{ className: string; onClick?: () => void }>,
+            Warning: Warning as React.ComponentType<{ className: string; onClick?: () => void }>,
+            CaretDown: CaretDown as React.ComponentType<{ className: string; onClick?: () => void }>,
+            Menu: Menu as React.ComponentType<{ className: string; onClick?: () => void }>,
+        },
+        forceUpdateTooltip: () => {
```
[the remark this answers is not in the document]
sqs: This is for extensions UI like the command palette and errors when adding an extension. Not for codeintellify. It accepts the icons as parameters so that the app using it can use native icons (GitHub’s loader for GitHub, Bitbucket’s loader for BB, etc.).

sqs:
@chrismwendt One other idea for getting the content of arbitrary files:

```
fetch('https://github.com/sourcegraph/browser-extensions/blob/master/README.md').then(x => x.text()).then(x => console.log(x))
```

seems to work, outputting the HTML for that blob page. You could then parse that HTML to obtain the raw file contents.

KattMingMing:
@sqs - I looked into this earlier as well and I think that might be the easiest option so that we do not need additional user steps. All the other alternatives required some sort of authorization step (e.g personal access token, OAuth application) which I think can be avoided with this implementation.

chrismwendt:
I replaced GitHub tokens with `document.querySelector('tbody').innerText` :+1:

`tree-list` + requests to other pages is a clever solution to xfiles/xcontent. I'll add support in cxp-js for that API then implement it here in browser-extensions.

app/backend/extensions.ts
```diff
@@ -0,0 +1,202 @@
...
+    // TODO fill in legit values
+    id: 'Client',
+    settingsURL: 'foo',
+    viewerCanAdminister: true,
+    __typename: 'Client',
+    displayName: 'me',
+}
+
+export const clientSettingsUpdates: Observable<string> = (() => {
+    const update = new ReplaySubject<string>(1)
```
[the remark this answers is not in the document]
chrismwendt: How do you convert an `Observable` into a `ReplaySubject`?

app/backend/lsp.tsx
```diff
@@ -302,7 +221,7 @@ export const fetchServerCapabilities = (pos: AbsoluteRepoLanguageFile): Observab
             for (const result of results) {
                 if (result && result.error) {
                     if (result.error.code === EMODENOTFOUND) {
-                        unsupportedModes.add(mode)
+                        unsupportedModes.add(pos.language)
```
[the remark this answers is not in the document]
chrismwendt: The term mode/language is in flux right now with the transition to CXP

app/util/polyfill.tsx
```diff
@@ -1,8 +1,17 @@
...
+//
+// tslint:disable-next-line:no-unbound-method
+const createObjectURL = window.URL ? window.URL.createObjectURL : null
+
 import { URL, URLSearchParams } from 'whatwg-url'
 
 const GLOBAL = global as any
 GLOBAL.URL = URL
 GLOBAL.URLSearchParams = URLSearchParams
+;(window.URL.createObjectURL as any) = createObjectURL
```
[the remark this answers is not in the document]
chrismwendt: I'm not sure, I grabbed it from https://sourcegraph.sgdev.org/github.com/sourcegraph/sourcegraph@b29b9e829f104ce22949bf0dfb11515a2da1ac11/-/blob/web/src/app.tsx#L10

I'll remove the cast and conditionally assign `createObjectURL`

extension/storage.ts
chrismwendt: I think this can be replaced by an `Observable` (and other occurrences of this).

PR #23, opened 2018-08-14
refactor: simplify feature flags data model

chrismwendt:
This adds the new tooltips feature flag as a checkbox in the Experimental features section of the options popup:

image

Testing plan:

- Toggle the new tooltips feature flag on/off and make sure it toggles between new and old tooltips

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle

### DIFF
--- a/app/cli/featureFlags.ts
+++ b/app/cli/featureFlags.ts
@@ -1,50 +1,40 @@
 import * as OmniCLI from 'omnicli'
 
-import storage from '../../extension/storage'
-import { featureFlagDefaults, FeatureFlags } from '../../extension/types'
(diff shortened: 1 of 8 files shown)


chrismwendt:
Ditching this because it's not necessary right now, and should be split into a few different PRs in the future:

- Make all feature flags booleans (right now they're `any`)
- Centralize feature flag metadata like title and default value (right now it's scattered in a few files)
- Unify the interface to toggling feature flags (all flags should be togglable using the `src :tff <flag>` command)
- Make feature flags subscribable with rxjs for real-time feedback

PR #25, opened 2018-08-16
build: upgrade to TypeScript 3.0.1

chrismwendt:


app/phabricator/backend.tsx
```diff
@@ -425,11 +425,14 @@ function convertConduitRepoToRepoDetails(repo: ConduitRepo): Promise<Phabricator
             // phabricator conduit API. Since we do not currently send the PHID with the Phabricator repository this a
             // backwards work around configuration setting to ensure mappings are correct. This logic currently exists
             // in the browser extension options menu.
-            const callsignMappings =
-                window.localStorage.PHABRICATOR_CALLSIGN_MAPPINGS || window.PHABRICATOR_CALLSIGN_MAPPINGS
+            type Mappings = { callsign: string; path: string }[]
+            const mappingsString = window.localStorage.getItem('PHABRICATOR_CALLSIGN_MAPPINGS')
+            const callsignMappings = mappingsString
+                ? (JSON.parse(mappingsString) as Mappings)
+                : window.PHABRICATOR_CALLSIGN_MAPPINGS || []
```
chrismwendt: @KattMingMing Can you double check this hunk? I think I preserved the existing semantics, but another pair of eyes would help. This might actually get completely obviated by your changes.

KattMingMing: This is actually served from the Phabricator extension (https://sourcegraph.com/github.com/sourcegraph/phabricator-extension/-/blob/rsrc/sourcegraph-config.js) so we do need it.

PR #31, opened 2018-08-16
feat: add a mock CXP implementation of LSP

chrismwendt:
This has no effect until you toggle on "Enable CXP".

This PR changes:

- Shows a mock CXP hover response when "Enable CXP" is on **and** you're on a single code file page on GitHub
- (internal) Adds a dependency on [cxp](https://yarnpkg.com/en/package/cxp)
- (internal) Adds a `SimpleCXPFns` interface that's implemented by both the `.api/xlang` endpoint and the mock CXP implementation

Testing plan:

- Toggle "Enable CXP" on/off and make sure hovers/defs/refs still work everywhere (it'll only show the mock CXP hover on single code file pages on GitHub)

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


app/github/inject.tsx
```diff
@@ -618,47 +619,65 @@ function injectBlobAnnotators(hoverifier: Hoverifier): void {
...
+
+    if (isSingleCodeFile) {
+        const lsp = useCXP ? lspViaCXP : lspViaAPIXlang
+        const hoverifier = createCodeIntelligenceContainer({ repoPath, lsp })
+        addBlobAnnotator(files[0] as HTMLElement, hoverifier, lsp)
+    } else {
+        const lsp = lspViaAPIXlang
+        const hoverifier = createCodeIntelligenceContainer({ repoPath, lsp })
+        for (const file of files) {
+            addBlobAnnotator(file as HTMLElement, hoverifier, lsp)
```
chrismwendt: @ijsnow This decides whether to use `.api/xlang` or CXP (based on whether this is one code file or multiple).

app/github/inject.tsx
```diff
@@ -87,9 +87,9 @@ export function injectGitHubApplication(marker: HTMLElement): void {
 function injectCodeIntelligence(): void {
     const { repoPath } = parseURL()
 
-    const hoverifier = createCodeIntelligenceContainer({ repoPath })
+    const hoverifier = createCodeIntelligenceContainer({ repoPath, simpleCXPFns: lspViaAPIXlang })
 
-    injectBlobAnnotators(hoverifier)
```
ijsnow: I originally didn't create the hoverifier in one place so I didn't have to pass it around everywhere but that led to bugs like this one https://github.com/sourcegraph/sourcegraph/issues/12620. With these changes being merged in, even without the feature flag enabled, we have two `hoverifiers` created at once which means two things can try to render the tooltip into the same mount. This is what caused the tooltip to hang around when navigating in the issue I linked to above. Not sure if its easily reproducible right now since this was merged but its definitely possible and we should always only have one instance of a hoverifier created at a time.

chrismwendt: Ah, I'll see if I can create only one

PR #32, opened 2018-08-16
Implement codeintellify support for Phabricator

ijsnow:
This PR is WIP. It only supports diffs in Phabricator and the styling isn't done. Just put this up so you can review what I've done so far since its a pretty big diff.


nicksnyder:
On Friday Isaac decided that it was not ready.

ijsnow:
@felixfbecker @chrismwendt this is ready to review

src/libs/code_intelligence/inject.tsx
```diff
@@ -0,0 +1,227 @@
...
+/** Information for adding code intelligence to code views on arbitrary code hosts. */
+export interface CodeHostInfo {
+    /**
+     * The name of the code host. This will be added as a className to the overlay mount.
+     */
+    name: string
+    /**
+     * The list of types of code views to try to annotate.
+     */
+    codeViews: CodeViewInfo[]
```
chrismwendt: How about `CodeView` instead of `CodeViewInfo`? "Info" to me sounds like plain data.

src/libs/code_intelligence/inject.tsx
```diff
@@ -0,0 +1,227 @@
...
+    filePath: string
+    /**
+     * The commit that the code view is at. If a `baseCommitID` is provided, this value is treated
+     * as the head commit ID.
+     */
+    commitID: string
+    /**
+     * The revision the code view is at. If a `baseRev` is provided, this value is treated as the head rev.
+     */
+    rev?: string
```
chrismwendt: Would it make sense to split `FileInfo` into two types, one for files and one for diffs? You might be able to avoid the optional properties then.

src/libs/code_intelligence/inject.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+    headHasFileContents?: boolean
+    baseHasFileContents?: boolean
+}
+
+export interface CodeViewInfo {
+    selector: string
+    dom: DOMFunctions
+    getToolbarMount?: (codeView: HTMLElement, part?: DiffPart) => HTMLElement
+    resolveFileInfo(codeView: HTMLElement): Observable<FileInfo>
```
chrismwendt: Is `resolveFileInfo` intended to emit once? I'd add some docs to each of these properties to describe their purpose.

src/libs/code_intelligence/inject.tsx
```diff
@@ -0,0 +1,227 @@
...
+    adjustPosition?: PositionAdjuster
+    toolbarButtonProps?: ButtonProps
+}
+
+function createCodeIntelligenceContainer(info: CodeHostInfo): { hoverifier: Hoverifier } {
```
chrismwendt: It wasn't obvious to me what a code intelligence container was until I saw that it mounts the hover tooltip. How about `mountHoverTooltip`, or something that conveys that this modifies the DOM and is related to tooltips? A doc block would help too, especially if this is intended to be called exactly once on the page.

src/libs/code_intelligence/inject.tsx
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    render(<HoverOverlayContainer />, overlayMount)
+
+    return { hoverifier }
+}
+
+export interface CodeView extends CodeViewInfo {
+    codeView: HTMLElement
+}
```
chrismwendt: Oh, I see there already exists a `CodeView` type. Here `codeView: HTMLElement` but above `codeViews: CodeViewInfo[]`, which seems inconsistent. A different word besides "Info" could improve the readability of this as well. As vague as it is, I think "Descriptor" would be a better choice, but there's probably some other words that convey a more specific meaning.

src/libs/code_intelligence/inject.tsx
```diff
@@ -0,0 +1,227 @@
...
+function findCodeViews(codeViewInfos: CodeViewInfo[]): Observable<CodeView> {
+    return new Observable<CodeView>(observer => {
+        for (const info of codeViewInfos) {
+            const elements = document.querySelectorAll<HTMLElement>(info.selector)
+            for (const codeView of elements) {
+                observer.next({ ...info, codeView })
+            }
+        }
+    })
+}
```
chrismwendt: Once you've found the code view, do you really need the `selector` anymore? Maybe it could be split into a separate type. I'm just trying to think of ways to restructure the data types such that it's easier to name them.

src/libs/code_intelligence/inject.tsx
```diff
@@ -0,0 +1,227 @@
...
+export interface CodeView extends CodeViewInfo {
+    codeView: HTMLElement
+}
+
+function findCodeViews(codeViewInfos: CodeViewInfo[]): Observable<CodeView> {
+    return new Observable<CodeView>(observer => {
+        for (const info of codeViewInfos) {
+            const elements = document.querySelectorAll<HTMLElement>(info.selector)
+            for (const codeView of elements) {
+                observer.next({ ...info, codeView })
```
chrismwendt: Since this is synchronous (it is, isn't it?), perhaps the return type could be simplified to `CodeView[]` instead of `Observable<CodeView>`.

src/libs/phabricator/backend.tsx
```diff
@@ -322,6 +324,43 @@ export function getRepoDetailsFromCallsign(callsign: string): Promise<Phabricato
     })
 }
 
+export function getRepoDetailsFromCallsignObservable(callsign: string): Observable<PhabricatorRepoDetails> {
```
chrismwendt: Unless there are other `getRepoDetails` functions, the `FromCallsignObservable` part is redundant with the type signature.

chrismwendt: Unless there are other `getRepoDetails` functions, `FromCallsignObservable` is redundant.

src/libs/phabricator/file_info.ts
```diff
@@ -0,0 +1,110 @@
...
+            const fetchingHeadFile = fetchBlobContentLines({
+                repoPath: info.headRepoPath,
+                filePath: info.filePath,
+                commitID: info.headCommitID,
+                rev: info.headRev,
+            })
+
+            return zip(fetchingBaseFile, fetchingHeadFile).pipe(
+                map(([baseFileContent, headFileContent]) => ({ ...info, baseFileContent, headFileContent }))
+            )
```
chrismwendt: Out of curiosity, why `zip` instead of perhaps `forkJoin`?

src/shared/util/featureFlags.ts
```diff
@@ -1,37 +1,69 @@
...
+const inPageFeatureFlags = createFeatureFlagStorage({
+    get: key => new Promise(resolve => resolve(!!localStorage.getItem(key))),
+    set: (key, val) =>
+        new Promise(resolve => {
+            localStorage.setItem(key, val.toString())
+            resolve(val)
+        }),
+})
+
+export const featureFlags: FeatureFlagsStorage = isInPage ? inPageFeatureFlags : browserExtensionFeatureFlags
```
chrismwendt: Nice

src/libs/code_intelligence/inject.tsx
ijsnow: To me, `CodeView` refers to the actual dom element and this is information about it.

src/libs/phabricator/file_info.ts
ijsnow: Tbh because I know what `zip` is and not `forkJoin`.

src/libs/code_intelligence/inject.tsx
ijsnow: That's not why `rev` is optional. No matter what, we're always looking at a file at a specific point in the git history. So it's "required" for single file views as well. The reason `rev` is optional because we don't always know the revision but we always need the commit. So we may know we're looking at commitID `123456` but not know that its actually pointing to rev `master`. However, if we don know it's master, we want to link to `sourcegraph.com/.../@master` rather than `sourcegraph.com/.../@123456`

src/libs/phabricator/backend.tsx
ijsnow: There are multiple `getRepoDetailsFrom<X>` in this file, thus the `FromCallsign` part. There is already a `getRepoDetailsFromCallsign` which is why I added the `Observable`. Once this isn't feature flagged and we're confident in the ability to remove stuff, I'll remove the old one and change this one's name.

cypress/integration/github_repository_spec.ts
```diff
@@ -29,8 +29,6 @@ describe('GitHub', () => {
 
     it('should() - (Unified PR) assert base tooltips are provided', () => {
         cy.visit('https://github.com/gorilla/mux/pull/328/files?diff=unified')
-        cy.contains(VIEW_FILE_BASE_BUTTON)
-        cy.contains(VIEW_FILE_HEAD_BUTTON)
```
ijsnow: @chrismwendt from my commit message:

> The rendering of the toolbar buttons and the support for code
intelligence is now less tighly coupled and happens in an asynchronous
way. This means that we shouldn't test in this order. There should
probably be other tests added for showing the buttons.

Let me know if you have any thoughts on this regarding merging this PR

felixfbecker:
Will have reviewed by Monday morning PST

ijsnow:
> Am I understanding correctly that this does add interfaces for code hosts, but only uses them for Phabricator, not yet for GitHub?

Yep! The original goal was just to use codeintellify in Phabricator but I wrote it in an abstract way so that we could use it with other code hosts once it's done.

> I have a couple of comments that I think should be resolved, other than that, great work!

Thanks! 😄

package-lock.json
```diff
@@ -2,6 +2,216 @@
   "requires": true,
```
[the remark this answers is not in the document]
ijsnow: 🤷‍♂️ I'll undo them

src/libs/code_intelligence/HoverOverlay.scss
```diff
@@ -0,0 +1,18 @@
+.hover-overlay__phabricator {
```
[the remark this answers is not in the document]
ijsnow: This is being used here https://github.com/sourcegraph/browser-extensions/blob/phab-codeintellify/src/libs/code_intelligence/inject.tsx#L119

I'm not sure what the correct way of doing this semantically or with BEM best practices but the reasoning is that I needed to modify some styles/add some styles and I don't want to modify them on the global level. The changes here would break things on github if they weren't contained.

This was one of the reasons for my growing opinion that plain CSS/SCSS that is injected into the page and bootstrap are bad for things like libraries and the browser extension where they'll be ran in many different environments with different styling. 

I know that it was nice when it "just worked" in github when pulling codeintellify out of the web app but thats where the fun stops. Not everywhere uses bootstrap and not everyone who does is on the same version.

src/shared/util/featureFlags.ts
```diff
@@ -1,37 +1,69 @@
...
+    disable: key => set(key, false),
+    isEnabled<K extends keyof FeatureFlags>(key: K): Promise<boolean> {
+        return get(key).then(val => !!val)
+    },
+    toggle<K extends keyof FeatureFlags>(key: K): Promise<FeatureFlags[K]> {
+        return get(key).then(val => set(key, !val))
+    },
+})
+
+function bextGet<K extends keyof FeatureFlags>(key: K): Promise<FeatureFlags[K]> {
```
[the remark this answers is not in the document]
ijsnow: browser extension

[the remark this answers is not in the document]
felixfbecker: Oh, that didn't come to my mind. I would consider renaming it to the long form, when referencing you get autocompletion anyway, but it can confuse people reading the code.

src/shared/util/featureFlags.ts
```diff
@@ -1,37 +1,69 @@
...
+const browserExtensionFeatureFlags = createFeatureFlagStorage({
+    get: bextGet,
+    set: bextSet,
+})
 
-/** Toggle boolean feature flags. */
-export function toggle<K extends keyof FeatureFlags>(key: K): Promise<FeatureFlags[K]> {
-    return get(key).then(val => set(key, !val))
-}
+const inPageFeatureFlags = createFeatureFlagStorage({
```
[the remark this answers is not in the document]
ijsnow: `localStorage` is different from `(chrome|browser).storage.(sync|local)`. Read more [https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage](here).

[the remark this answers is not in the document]
felixfbecker: I understand that, but why do we need two places to store feature flags in?

src/libs/phabricator/code_views.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        const mountLocation = findMountLocation(file, part)
+        mountLocation.appendChild(mount)
+
+        return mount
+    }
+}
+
+// Gets the actual text content we care about and returns the number of characters we have stripped
+// so that we can adjust accordingly.
```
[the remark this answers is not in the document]
felixfbecker: Yeah, I understood this from reading code and other comments, but it would be better if it was explained in the docblock of the function.

src/libs/code_intelligence/inject.tsx
ijsnow: I hear both of your complaints and am willing to change this to make it easier to understand but can you give me a concrete example of how you'd like it to be? It makes sense to me with the context I had while working on this PR and can't think of a clean and easy way to break this apart.

One thing I'd like to avoid is building types out of the `&` operator. I get why it's helpful but there is nothing more disappointing to me than clicking on "Go to definition" and seeing that I have to now go back and forth from like 10 different types to see what something actually looks like.

src/libs/code_intelligence/inject.tsx
```diff
@@ -0,0 +1,246 @@
...
+    /** Resolves the file info for a given code view. It returns an observable because some code hosts need to resolve this asynchronously. The observable should only emit once. */
+    resolveFileInfo(codeView: HTMLElement): Observable<FileInfo>
+    /** In some situations, we need to be able to adjust the position going into and coming out of codeintellify. For example, Phabricator converts tabs to spaces in it's DOM. */
+    adjustPosition?: PositionAdjuster
+    /** Props for styling the buttons in the `CodeViewToolbar`. */
+    toolbarButtonProps?: ButtonProps
+}
+
+/**
+ * initCodeIntelligence prepares the page for code intelligence. It creates the hoverifier, injects and mounts the hover overlay
```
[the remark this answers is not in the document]
ijsnow: Wish that the tooling supported it 😉

src/shared/util/featureFlags.ts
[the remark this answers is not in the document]
ijsnow: When the code is ran as a browser extension it'll use the browser extension storage and when its ran as a part of the page's scripts(phabricator extension), we'll use `localStorage` because the browser extension storage doesn't exist.

src/libs/code_intelligence/inject.tsx
felixfbecker: Can there just be two interfaces, one for the base case, and one for diffs, where the diff interface extends the base interface?

src/shared/util/featureFlags.ts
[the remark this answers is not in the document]
felixfbecker: Ah, got it. Thanks.

src/shared/util/featureFlags.ts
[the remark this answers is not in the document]
felixfbecker: This would be a candidate for renaming

src/shared/util/featureFlags.ts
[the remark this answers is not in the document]
felixfbecker: Got it, thanks. This would be a candidate for adding a comment

src/libs/code_intelligence/HoverOverlay.scss
```diff
@@ -0,0 +1,18 @@
+.hover-overlay-mount__phabricator {
```
felixfbecker: Should be `--phabricator` (dashes not underscore)

ijsnow: Everywhere else in our codebases(that I've seen), use underscore. What is the difference between underscore and dashes in BEM?

felixfbecker: They are less frequent than elements, but they do appear: https://sourcegraph.sgdev.org/search?q=f:%5C.scss%24+%26--

Differences are best explained by http://getbem.com/naming/

PR #35, opened 2018-08-16
refactor: reuse graphql types from extensions-client-common

chrismwendt:
Here's `types/gqlschema.d.ts` (hard to see in the massive diff):

```typescript
import * as GQL from '@sourcegraph/extensions-client-common/lib/schema/graphqlschema'

export { GQL }
```

This is not considered best practice. The goal here is to avoid copy-pasting.


felixfbecker:
Ah, I see, it was just moved to typings/.
But how does this save copy+pasting/what was copy pasted? The GQL types are generated with npm run graphql from the live introspection endpoint

chrismwendt:
My bad, I didn't notice the `npm run graphql` command. The GQL types exported from e-c-c conflicted with the ones that were in this repo.

felixfbecker:
If they conflict, that could be an actual type incompatibility that we would not want to hide - i.e. if the version installed expects the returned objects of an old GQL API version, but our new types here dont provide that, that should give the hint that ecc needs to be updated. I.e. I think every repo should have their own typings

PR #38, opened 2018-08-17
feat(cxp): add options page with CXP extensions list

chrismwendt:
This PR changes:

- Adds an options page under the "Enable CXP" feature flag (right click on the icon, select Options)

Testing plan:

- Check "Enable CXP"
- Open the options page

- You should see CXP extensions from your Primary Sourcegraph instance
- Toggling extensions on/off should change their associated value in the client settings at the bottom of the page to true/false
- Removing an extension should remove it from the client settings
- Adding an extension should add it to client settings

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle

cc @KattMingMing Take note of `cxp.html` and `cxp.tsx` - these will get replaced by or embedded into your new options UI.

TODO

- [x] Don't show user/org/site options because this is for client settings only
- ~[ ] Link to the upstream extension registry when the user clicks on an extension~ easier accomplished by changing the GraphQL API of Sourcegraph


app/extensions-client-common.scss
```diff
@@ -0,0 +1,111 @@
...
+        font-weight: 600;
+    }
+    h5 {
+        font-size: 12px;
+        letter-spacing: 1px;
+        font-weight: 500;
+        text-transform: uppercase;
+        margin: 0;
+    }
+}
```
chrismwendt: I wrapped the selectors that were likely to conflict with GitHub's CSS and I double checked a few pages to gain confidence that the non-wrapped selectors didn't break anything.

app/extensions-client-common.scss
```diff
@@ -0,0 +1,111 @@
...
+.nav img.icon-inline {
+    // TODO find a better way to scale icons that is not proportional to the font size
+    // 14px font size -> 20px icon size
+    // stylelint-disable-next-line declaration-property-unit-whitelist
+    width: (20em / 14);
+    // stylelint-disable-next-line declaration-property-unit-whitelist
+    height: (20em / 14);
+    // Match color of the text
+    fill: currentColor;
+}
```
chrismwendt: This is a relatively minimal set of CSS necessary for the extensions list to look similar to Sourcegraph.com.

PR #45, opened 2018-08-17
build: fetch GQL schema from endpoint

felixfbecker:


felixfbecker:
So, there are now some compile errors - in theory they should go away if ecc gets updated to the latest schema.

chrismwendt:

```
cd browser-extensions

cd ../sourcegraph
# and fixup merge conflicts
```

PR #46, opened 2018-08-17
feat(options): add new full page options

KattMingMing:
This PR changes:

- Adds feature flagged full page options menu
- Reduces `serverUrls` to 1 url and https://sourcegraph.com
- Adds empty CXP section for options page
- Fetches configuration from server instances

Testing plan:

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


app/backend/graphql.tsx
```diff
@@ -146,7 +160,9 @@ export function queryGraphQL(ctx: RequestContext, query: string, variables: any
  * @return Observable That emits the result or errors if the HTTP request failed
  */
 export function mutateGraphQL(ctx: RequestContext, mutation: string, variables: any = {}): Observable<MutationResult> {
-    return requestGraphQL(ctx, mutation, variables, serverUrls) as Observable<MutationResult>
+    return requestGraphQL(ctx, mutation, variables, [sourcegraphUrl, 'https://sourcegraph.com']) as Observable<
```
felixfbecker: Could you explain this change?

KattMingMing: I'd like to remove `serverUrls` as a concept and only use one `sourcegraphUrl`. I think the better place to do this is to have `requestGraphQL` use the Sourcegraph URL and then in the retry logic use `https://sourcegraph.com` instead of continuing to pass an array of URLs.

app/components/options/OptionsDashboard.tsx
```diff
@@ -0,0 +1,24 @@
...
+export class OptionsDashboard extends React.Component<any, {}> {
+    public render(): JSX.Element {
+        return (
+            <HashRouter>
+                <div className="site-admin-area area">
+                    <OptionsPageSidebar className="area__sidebar" />
+                    <div className="area__content">
+                        <Switch>
+                            <Route path="/" component={OptionsConfiguration} exact={true} />
+                            <Route path="/extensions" component={CXPExtensionRegistry} exact={true} />
```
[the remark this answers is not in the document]
KattMingMing: Yep - I moved this to the sidebar not the actual route handling

PR #50, opened 2018-08-17
feat(cxp): add UI components for command contributions

chrismwendt:
image


app/github/inject.tsx
```diff
@@ -201,6 +209,42 @@ function hideFileTree(): void {
...
+function injectCXPGlobalComponents({
+    cxpController,
+    extensionsContextController,
+}: {
+    cxpController: CXPController<ExtensionWithManifest, ConfigurationCascade<ConfigurationSubject, Settings>>
+    extensionsContextController: Controller<ConfigurationSubject, Settings>
+}): void {
+    const statusElem = document.createElement('div')
+    statusElem.id = 'cxp-global'
+    statusElem.style.position = 'fixed'
```
KattMingMing: Can you move the inline styles to a classname

app/github/inject.tsx
```diff
@@ -201,6 +209,42 @@ function hideFileTree(): void {
...
 
+function injectCXPGlobalComponents({
+    cxpController,
+    extensionsContextController,
+}: {
+    cxpController: CXPController<ExtensionWithManifest, ConfigurationCascade<ConfigurationSubject, Settings>>
+    extensionsContextController: Controller<ConfigurationSubject, Settings>
+}): void {
+    const statusElem = document.createElement('div')
+    statusElem.id = 'cxp-global'
```
KattMingMing: Where is this ID used?

KattMingMing:
@chrismwendt  can we get the command input to extend the full width of the popup (with some horizontal padding)

app/github/inject.tsx
chrismwendt: Moved to a classname (not pushed yet)

app/github/inject.tsx
chrismwendt: I don't think it's used at all - removed

chrismwendt:
Yes, I'm tweaking the CSS right now

PR #57, opened 2018-08-19
build: speed up webpack build using HardSource for caching

chrismwendt:
https://github.com/mzgoddard/hard-source-webpack-plugin

HardSource caches modules in `.cache/` and cuts down the build time:

- Initial build: 55s down to 15s 🚤💨 
- Incremental build: 10s down to 7s (based on light empirical evidence)

Here's what the output looks like 👀:

```
[hardsource:57550c93] Using 69 MB of disk space.
[hardsource:57550c93] Tracking node dependencies with: package-lock.json.
[hardsource:57550c93] Reading from cache 57550c93...
[16:36:53] › …  awaiting  Compiling...
...
```

Unfortunately, I ran into the same issue reported here ☹️ https://github.com/mzgoddard/hard-source-webpack-plugin/issues/416 and I avoided it by removing node-sass-import-once and dropping the `.css` extension on some SASS imports. What's left are some deprecation warnings that will be dropped soon https://github.com/sass/node-sass/issues/2362#issuecomment-401964451 These deprecation warnings can be ignored 🙈:

```
DEPRECATION WARNING on line 4, column 8 of /Users/chrismwendt/sourcegraph/browser-extensions/app/extensions-client-common.scss:
Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass.
Use a custom importer to maintain this behaviour. Check your implementations documentation on how to create a custom importer.
...
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ npm-debug.log.*
 # Cypress testing. These will get downloaded as artifacts by buildkite.
 cypress/screenshots/
 cypress/vidoes/
+
+.cache/
--- a/app/app.scss
+++ b/app/app.scss
@@ -89,5 +89,5 @@ $theme-colors-light: (
 @import './components/symbols';
 @import './components/codeIntelStatusIndicator';
 @import 'node_modules/@sourcegraph/codeintellify/lib/HoverOverlay.scss';
-@import 'node_modules/@sourcegraph/react-loading-spinner/lib/LoadingSpinner.css';
+@import 'node_modules/@sourcegraph/react-loading-spinner/lib/LoadingSpinner';
 @import './extensions-client-common';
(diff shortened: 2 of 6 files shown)


felixfbecker:
This is awesome!

> and I avoided it by removing node-sass-import-once 

sass-import-once is there for a reason though - otherwise Sass imports don't work like TS imports where a module is only included once. Every file or library that imports the same file will actually bring in that file another time into the bundle, but every file or dependency should be able to e.g. declare a dependency on bootstrap without bringing Bootstrap multiple times. This is not just a size concern, as the order in which CSS is defined matters.

chrismwendt:
I see, the removal of node-sass-import-once breaks the CSS:

image

This shouldn't be a problem if/when we switch from SASS to some CSS-in-JS solution (e.g. aphrodite). Unless there's a way to fix this sooner, let's table HardSource until then.

felixfbecker:
I don't know, I would not think so. I would recommend you post a comment on that thread saying that you identified this only happens with sass-import-once, because I don't see it mentioned in the thread.

The build of the browser-extension is not very optimized. I think there are other ways we can speed it up we can also try.

PR #60, opened 2018-08-20
build(webpack): do not poll

chrismwendt:
I noticed `node` taking 50% CPU consistently while `npm run dev` was running.

I removed polling from webpack and that seems to have fixed it. Compilation still gets triggered on file save.

Does anyone depend on this? cc @ijsnow @KattMingMing @felixfbecker


chrismwendt:
@KattMingMing said this should work too :+1:

ijsnow:
I wonder if this is how it was picking up changes from `node_modules` and recompiling. Either way, that's not all that important.

PR #72, opened 2018-08-21
fix: missing configure Sourcegraph buttons with empty URL

KattMingMing:
This PR resolves https://github.com/sourcegraph/sourcegraph/issues/12930:

- Checks for a Sourcegraph URL before fetching to fix requests going to GitHub.com.
- Removes unnecessary safari check (safari is not currently maintained)
- Renders **Configure Sourcegraph** button in page headers when no URL is set or only https://sourcegraph.com is set and user is viewing a private repository.
- Clicking **Configure Sourcegraph** with options page feature flag enabled opens new options page in a new tab.

Testing plan:

- Remove Sourcegraph URL from the feature flagged options page.
- Go to a private repository
- **Configure Sourcegraph** button appears in page headers
- Clicking on Configure Sourcegraph goes to new options page with correct error message.

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


chrome/extension/inject.tsx
```diff
@@ -37,15 +37,6 @@ function injectApplication(): void {
             return
         }
 
-        // This has a default value so it should always be defined.
-        // In safari, if the storage hasn't been initialized yet, this
-        // will be undefined so we want to return early.
-        // When safari's storage is initialized, we fire a custom event that will
-        // re-run this function.
-        if (!items.sourcegraphURL) {
```
ijsnow: Hopefully we don't forget this in the scenario we decide to support Safari again

app/backend/graphql.tsx
```diff
@@ -33,7 +33,7 @@ function requestGraphQL(
     authError?: AuthRequiredError
 ): Observable<GQL.IGraphQLResponseRoot> {
     // Check if it's a private repo - if so don't make a request to Sourcegraph.com.
-    if (isPrivateRepository() && url === 'https://sourcegraph.com') {
+    if ((isPrivateRepository() && url === 'https://sourcegraph.com') || !url) {
```
felixfbecker: `url` is defined as type `string`, and it being named `url`, I would never expect it to be a non-URL, including an empty string. A sanity check here is fine, but this probably should be fixed at a different layer? I.e. where you configure the URL, it should default it back to `https://sourcegraph.com` if set to an empty URL, and it should validate that it is a proper URL, so invalid strings aren't even passed here.

A way to avoid this sanity check here and make it type safe would be to pass a `URL` instance instead of a string, which sounds like a good idea given we met this bug. It's easy to convert them back and forth (`new URL(str)` and `url.href`), and the constructor will validate and throw if the URL is invalid. Any method can then rely that a passed `URL` instance is valid.

extension/runtime.ts
```diff
@@ -56,9 +56,19 @@ export const getContentScripts = () => {
...
+/**
+ * openOptionsPage opens the options.html page. This must be called from the background script.
+ * @param callback Called when the options page is opened.
+ */
 export const openOptionsPage = (callback?: () => void): void => {
-    if (chrome && chrome.runtime && chrome.runtime.openOptionsPage) {
-        chrome.runtime.openOptionsPage(callback)
+    if (chrome && chrome.runtime) {
+        const extensionUrl = chrome.runtime.getURL('').replace(/\/$/, '')
+        if (window.location.origin !== extensionUrl) {
```
ijsnow: Could this could be done with [these values](https://sourcegraph.sgdev.org/github.com/sourcegraph/browser-extensions/-/blob/app/context.ts#L42-44)?

app/components/options/ConnectionCard.tsx
```diff
@@ -60,6 +64,23 @@ export class ConnectionCard extends React.Component<Props, State> {
                 </div>
             )
         }
+        if (!sourcegraphUrl) {
+            return (
+                <div className="pt-2">
+                    <Alert color="danger">
+                        Error: missing Server URL. For code intelligence on public repositories add
+                        https://sourcegraph.com. For hover tooltips, jump to definition, and find references on public
```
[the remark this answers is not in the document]
KattMingMing: I'll update it so that users must have at least 1 URL.

chrome/extension/background.tsx
```diff
@@ -247,6 +247,10 @@ runtime.onMessage((message, _, cb) => {
 
         case 'setBadgeText':
             browserAction.setBadgeText({ text: message.payload })
+            return
+        case 'openOptionsPage':
```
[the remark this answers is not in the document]
KattMingMing: I added types for message instead of just having it as `string`

extension/runtime.ts
```diff
@@ -1,11 +1,22 @@
...
+        | 'getIdentity'
+        | 'setEnterpriseUrl'
+        | 'setSourcegraphUrl'
+        | 'removeEnterpriseUrl'
+        | 'insertCSS'
+        | 'setBadgeText'
+        | 'openOptionsPage'
+        | 'fetched-files'
+        | 'repo-closed'
     payload?: any
```
felixfbecker: Not blocking, but you can make `payload` also type safe by defining things like this:
```ts
type Message = SetIdentityMessage | GetIdentityMessage | ...

interface SetIdentityMessage {
    type: 'setIdentity'
    payload: { identity: string }
}
interface GetIdentityMessage {
    type: 'getIdentity'
}
...
```
TypeScript can then switch between the interfaces on the `type` property

KattMingMing: I'll follow up with this in another PR

Issue #21, opened 2018-08-14
No indication that C# support is experimental on the GitHub extension


felixfbecker:
_From @JoshuaKGoldberg on August 14, 2018 17:33_

Following #39 - as a user, it is very confusing to me that C# files randomly "don't work" (e.g. hover info takes a while to load, then has no content).

Is there some kind of UI that can explain that C# support is experimental? If not, should it be mentioned somewhere?

Also, are there plans (roadmaps?) for making C# support... not-experimental?

_Copied from original issue: sourcegraph/codeintellify#40_

felixfbecker:
@JoshuaKGoldberg we show a code intelligence status indicator in the top right:

<img width="189" alt="image" src="https://user-images.githubusercontent.com/10532611/44108283-38c56a54-9ffa-11e8-89f8-23445286d20f.png">

<img width="224" alt="image" src="https://user-images.githubusercontent.com/10532611/44108295-401c2b08-9ffa-11e8-86fa-bcc17fbea532.png">

Maybe we should make it more prominent.

felixfbecker:
_From @JoshuaKGoldberg on August 14, 2018 17:49_

Ah, I should be more specific - using the GitHub extension. Updated the title.

felixfbecker:
Ah, the same indicator should be on GitHub:

<img width="510" alt="image" src="https://user-images.githubusercontent.com/10532611/44109013-48e53ad4-9ffc-11e8-9a74-edfff8742a2c.png">

<img width="236" alt="image" src="https://user-images.githubusercontent.com/10532611/44109028-4def5da2-9ffc-11e8-9e73-5cb98d1ff32b.png">

This is for Go, for some reason the extension fails to retrieve the status, while sourcegraph.com works.

I'll move this issue to the browser extension

felixfbecker:
@KattMingMing I'm assigning this to you since you introduced the codeintel status indicator to the browser ext afaik

PR #75, opened 2018-08-21
fix(codeintel): show experimental language server warning

KattMingMing:
This PR resolves https://github.com/sourcegraph/browser-extensions/issues/21:

- Fixes a condition where the server was not returning back the correct experimental flags. Also note that sometimes the lang server takes a while to fetch information from (I checked on sourcegraph.com and the behavior is the same)

- Removes `textDocument/didOpen` and replaces with `initialize`

Testing plan:

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


app/components/CodeIntelStatusIndicator.tsx
```diff
@@ -46,32 +46,40 @@ const CapabilityStatus: React.StatelessComponent<{ label: string; provided: bool
...
                 ),
-                catchError(err => (err.code === EMODENOTFOUND ? [null] : [asError(err)]))
+                catchError(err => [asError(err)])
             )
         }),
         map(langServerOrError => ({ langServerOrError }))
     )
 
-interface CodeIntelStatusIndicatorProps extends AbsoluteRepoFile {
+interface CodeIntelStatusIndicatorProps extends AbsoluteRepoFile, ModeSpec {
```
ijsnow: I think I remember removing `mode` as a prop because we can just compute it from the `filePath` in here. What's the added benefit of having it passed in as a prop? It seems to me like we're just passing redundant information around.

app/components/CodeIntelStatusIndicator.tsx
```diff
@@ -93,7 +101,7 @@ export class CodeIntelStatusIndicator extends React.Component<
     private componentUpdates = new Subject<CodeIntelStatusIndicatorProps>()
     private subscription = this.componentUpdates
         .pipe(propsToStateUpdate)
-        .subscribe(stateUpdate => this.setState(stateUpdate))
+        .subscribe(stateUpdate => this.setState(() => ({ ...stateUpdate })))
```
ijsnow: Was this necessary?

app/backend/lsp.tsx
```diff
@@ -259,11 +252,29 @@ const fetchServerCapabilities = (pos: AbsoluteRepoLanguageFile): Observable<Serv
...
+            [
+                {
+                    id: 0,
+                    method: 'initialize',
+                    params: {
+                        rootUri: `git://${pos.repoPath}?${pos.commitID}`,
+                        mode,
+                        initializationOptions: { mode },
+                    },
+                },
```
chrismwendt: Was the issue with `wrapLSP` that `mode` was only present under `initializationOptions` and not directly under `params`? What does this solve?

app/components/CodeIntelStatusIndicator.tsx
```diff
@@ -46,32 +46,40 @@ const CapabilityStatus: React.StatelessComponent<{ label: string; provided: bool
...
-                        experimental: (langServer && langServer.experimental) || undefined,
-                        capabilities,
-                    })
+                    ([langServer, capabilities]): LangServer | null =>
+                        langServer && {
+                            displayName: langServer.displayName,
+                            homepageURL: langServer.homepageURL || undefined,
+                            issuesURL: langServer.issuesURL || undefined,
+                            experimental: langServer.experimental,
+                            capabilities,
```
chrismwendt: I'm not understanding the connection between this `experimental` flag and the `mode` change in this PR. Could you elucidate that connection?

app/components/CodeIntelStatusIndicator.tsx
KattMingMing: DidOpen didn't need to get called to resolve information about the lang server health / status - it will however surface errors related to that file if there are file level errors (which would also been seen when hovering)

chrismwendt: I've spent a lot of time shuffling the mode around too - it's a pretty annoying piece.

felixfbecker: Wasn't the mode separated out by @sqs to pave the way towards multiple langservers per file or something?

app/backend/lsp.tsx
felixfbecker: It seems like now we're sending `mode` in two ways? Shouldn't there be only one way?

felixfbecker:
@ijsnow @chrismwendt `didOpen` was literally just sent back in the day because the lsp-proxy had a bug in it that made it panic if the amount of messages was not exactly 4. It is fixed now, so we should remove `didOpen`.
@KattMingMing It also cannot surface any file level errors afaik because it is a notification, and notifications don't have responses (so also no error responses).

I agree with @chrismwendt though that it is not clear what the actual _cause_ of the bug was and how that cause is being fixed here.

app/backend/lsp.tsx
KattMingMing: Yes - I forgot to revert some of my changes before committing ha

KattMingMing:
@felixfbecker - I've updated this just to update the LSP call which fixes the issues. The "file" related issue was for didOpen would return errors into the code intel status indicator like:
```
Could not find /Users/administrator/cache/e173f1da-272a-4554-8b9f-e4f8257ee299/Penc/CGRectExtension.swift in the workspace.
``` that's resolved by only calling initialize for langServer information.

I just checked with the latest production version and it's returning the experimentalLanguages flag for C# correctly.

![screenshot 2018-08-21 13 09 43](https://user-images.githubusercontent.com/4897310/44426501-a84fe400-a543-11e8-8fbb-f814631fdba0.png)

felixfbecker:
Okay, that's then actually also an LSP proxy bug. Notifications cannot return responses (just like `exit` has none).

KattMingMing:
@felixfbecker - good to merge this clean up to remove `didOpen`?

felixfbecker:
It seems to still send `mode` twice

PR #80, opened 2018-08-22
fix: private server instance cors error

KattMingMing:
This PR changes:

.

Testing plan:

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


app/backend/lsp.tsx
```diff
@@ -247,11 +247,7 @@ const fetchServerCapabilities = (pos: AbsoluteRepoLanguageFile): Observable<Serv
     return ajax({
         method: 'POST',
         url: `${url}/.api/xlang/initialize`,
-        headers: {
-            ...getHeaders(),
-            Accept: 'application/json',
-            'Content-Type': 'application/json',
-        },
+        headers: getHeaders(),
```
felixfbecker: It might be good to add a comment here so this doesn't happen again

PR #81, opened 2018-08-22
feat(code-intel): add enable code intel to status indicator

KattMingMing:
This PR changes:

Resolves https://github.com/sourcegraph/sourcegraph/issues/12730

Testing plan:

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


nicksnyder:
screenshot?

app/components/CodeIntelStatusIndicator.tsx
```diff
@@ -173,6 +179,41 @@ export class CodeIntelStatusIndicator extends React.Component<
...
+                        <div>
+                            <Button
+                                color="primary"
+                                size="sm"
+                                className="mt-1"
+                                href={`${sourcegraphUrl}/site-admin/code-intelligence`}
+                                target="_blank"
+                            >
+                                Enable
+                            </Button>
```
chrismwendt: Does this still show up when the user is not an admin? Maybe the message could be different for admins vs non-admins.

KattMingMing: Good catch - updating now

app/components/CodeIntelStatusIndicator.tsx
```diff
@@ -173,6 +175,52 @@ export class CodeIntelStatusIndicator extends React.Component<
         this.setState(({ enabled }) => ({ enabled: !enabled }))
     }
 
+    private handleNoCodeIntelligence = (language?: string) => {
+        const { site, userIsSiteAdmin } = this.props
+        if (site) {
+            const { hasCodeIntelligence } = site
+            if (!hasCodeIntelligence) {
+                if (userIsSiteAdmin) {
```
nicksnyder: shouldn't this be inverted?

KattMingMing: Why yes it should - I was taking a screenshot and forgot to put this back

PR #87, opened 2018-08-23
feat(cxp): pass btn styles and update ecc

KattMingMing:
This PR resolves https://github.com/sourcegraph/sourcegraph/issues/12971:

Depends on https://github.com/sourcegraph/extensions-client-common/commit/3b96f058c9df1a6958da5745dde2b58e6794f02f being released and upgrading extension client common version.

- Passes action styles to `ActionsNavItems` to render toolbar buttons as buttons styled to match the code host.

screenshot 2018-08-23 16 27 47

Testing plan:

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


app/components/options/CXPExtensionRegistry.tsx
```diff
@@ -9,12 +9,19 @@ import { RouteComponentProps } from 'react-router-dom'
...
 
 interface OptionsPageProps extends RouteComponentProps<{}> {}
 interface OptionsPageState extends ConfigurationCascadeProps<ConfigurationSubject, Settings> {}
 
 const extensionsContextController = createExtensionsContextController()
 
+/** A fallback configuration subject that can be constructed synchronously at initialization time. */
+const SITE_SUBJECT: Pick<GQL.IConfigurationSubject, 'id' | 'viewerCanAdminister'> = {
+    id: 'CLIENT',
+    viewerCanAdminister: false,
```
KattMingMing: @chrismwendt / @sqs  - When I upgraded extensions client common I needed to stub this object to pass as the new subject. Should `viewerCanAdminister` be true or false? If it's false then the user cannot toggle the extension from the options page.

chrismwendt: Use `id: 'Client'` instead of `CLIENT` for consistency with https://github.com/sourcegraph/browser-extensions/blob/03e382b5a832a0f3c8b3c8705f072f3c2f43ab53/app/backend/extensions.ts#L35 (I'm not sure what the consequences of inconsistency are, maybe the setting changes wouldn't actually get applied)

`viewerCanAdminister` should be true because in the browser extension we always have read/write access to `browser.storage`, where the client settings are stored.

app/components/options/CXPExtensionRegistry.tsx
```diff
@@ -40,7 +47,7 @@ export class CXPExtensionRegistry extends React.Component<OptionsPageProps, Opti
                 </div>
                 <ExtensionsList
                     {...this.props}
-                    subject={'Client'}
+                    subject={SITE_SUBJECT}
```
chrismwendt: `CLIENT_SUBJECT` would be more accurate

app/github/inject.tsx
```diff
@@ -313,15 +314,15 @@ function injectCXPGlobalComponents({
...
+    cxpController: CXPController<ConfigurationSubject, Settings>
     extensionsContextController: Controller<ConfigurationSubject, Settings>
 }): void {
     const statusElem = document.createElement('div')
     statusElem.className = 'cxp-global'
     document.body.appendChild(statusElem)
     render(
         <ExtensionStatusPopover
-            cxpController={cxpController}
+            cxpController={cxpController as any}
```
chrismwendt: `as any` shouldn't be necessary - were there type errors?

KattMingMing: That was resolved when we all chatted about the ecc changes. Shouldn’t be necessary

PR #94, opened 2018-08-24
feat: increase the height of the client settings

chrismwendt:
Before:

image

After:

image


app/options.scss
```diff
@@ -130,7 +130,8 @@
             font-size: 14px;
             border-color: #e4e9f1;
             margin-bottom: 1rem;
-            height: 18rem;
+            // !important overrides Bootstrap's `height: auto`
+            height: 18rem !important;
```
chrismwendt: This is pretty hacky - any other ides?

PR #98, opened 2018-08-24
feat(review-board): add review board support

KattMingMing:
This PR changes:

- Uses codeintellify to add support for Review Board which is currently being beta.

Testing plan:

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


app/review-board/backend/fetch.ts
```diff
@@ -0,0 +1,19 @@
+import { from, Observable } from 'rxjs'
+import { ReviewBoardRepository } from '../util'
+
+function getRepository(repoID: number): Promise<ReviewBoardRepository> {
+    return new Promise((resolve, reject) => {
```
felixfbecker: You shouldn't use the Promise constructor here, `fetch` already returns a Promise.

app/review-board/dom/util.tsx
```diff
@@ -0,0 +1,119 @@
...
+ */
+export const diffDomFunctions: DOMFunctions = {
+    getCodeElementFromTarget: target => {
+        const codeCell = getCodeCellFromTarget(target)
+        return codeCell && getBlobCodeInner(codeCell)
+    },
+    getCodeElementFromLineNumber: () => null,
+    getLineNumberFromCodeElement,
+    getDiffCodePart,
+}
```
felixfbecker: I would propose to not have these in a `util` file, but in a file `dom_functions` like we do for GitHub.

app/review-board/inject.tsx
```diff
@@ -0,0 +1,197 @@
...
+    )
+}
+
+function injectAnnotators(): void {
+    if (!reviewBoardState) {
+        return
+    }
+    const { repository, reviewRequest } = reviewBoardState
+    getRepositoryFromReviewBoardAPI(repository.id).subscribe(rbRepository => {
+        resolveRepoPathFromName(rbRepository).subscribe(
```
felixfbecker: Don't nest subscribe, instead use switchMap

app/review-board/util.ts
```diff
@@ -0,0 +1,114 @@
...
+            branch,
+            commitID,
+            extraData,
+            id,
+            lastUpdated,
+            localSitePrefix,
+            repository,
+            reviewURL,
+            state,
+        }
```
felixfbecker: Is your goal here just to clone the object? Why not use `{ ...page.reviewRequest.attributes }`?

app/review-board/util.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    document.dispatchEvent(
+        new CustomEvent('REVIEW_BOARD_LOADED', {
```
felixfbecker: Use the constant?

app/review-board/util.ts
```diff
@@ -0,0 +1,114 @@
...
+            detail: {
+                repository: page.reviewRequest!.attributes.repository,
+                reviewRequest,
+            } as ReviewBoardState,
```
felixfbecker: Casting hides errors, it's better to declare it as a variable first.

types/globals/index.d.ts
```diff
@@ -15,6 +15,12 @@ interface Window {
     }
     // Bitbucket has a global require function on the DOM that we rely on to get the current Bitbucket state.
     require: any
+    // Reviewboard has a global RB function on that DOM that we rely on to get ReviewBoard state.
+    RB: {
+        PageManager: {
+            getPage: () => any
```
felixfbecker: Can this be typed instead of `any`? It will be hard for people to maintain if they don't know what this returns

chrismwendt:

```
cd browser-extensions

cd ../sourcegraph
# and fixup merge conflicts
```

PR #102, opened 2018-08-26
build: fix type errors for linking dependencies

chrismwendt:
This fixes some type errors when linking dependencies (`npm link` or `yarn link`).

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,9 @@
 {
   "compilerOptions": {
+    "baseUrl": "",
+    "paths": {
+      "rxjs": ["node_modules/rxjs"]
+    },
     "outDir": "./out",
     "target": "es2015",
     "module": "commonjs",
--- a/app/tree/Tabs.tsx
+++ b/app/tree/Tabs.tsx
@@ -69,7 +69,7 @@ interface TabsProps<T extends string> {
      */
     tabBarEndFragment?: React.ReactFragment
 
-    children: undefined | React.ReactElement<{ key: T }> | (undefined | React.ReactElement<{ key: T }>)[]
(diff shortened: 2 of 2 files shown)


tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
   "compilerOptions": {
+    "baseUrl": "",
+    "paths": {
+      "rxjs": ["node_modules/rxjs"]
+    },
```
chrismwendt: I ran into https://github.com/Microsoft/typescript/issues/6496 and adding `rxjs` to `paths` fixed the issue.

chrismwendt: While attempting to replace `Observable` with `Subscribable` in e-c-c, I ran into an issue where e-c-c accesses `.pipe()` on some input `Observable`s in [clientCommands.ts](https://sourcegraph.com/github.com/sourcegraph/extensions-client-common@d0a33acadc3b9823ea896fa8c98aa4903321ccbd/-/blob/src/client/clientCommands.ts#L64:9) (there are other instances), but `Subscribable`s don't have a `pipe()` function. Hoping to find another interface to use, I grepped through rxjs's type definitions but I didn't see any interface that has `pipe()` (only the `Observable` class).

Any other ideas? 💭

felixfbecker: Yeah, you need to pass the `Subscribable` to `from()` to convert it to a full `Observable`.

chrismwendt: Ah, got it, thanks 🙇

felixfbecker:
Can this be closed or are there still type errors?

chrismwendt:
There are still some type errors, but I'll try resolving them without `baseUrl`/`paths` in another PR.

PR #109, opened 2018-08-30
fix(permissions): add context menu to enable domain

KattMingMing:
This PR changes:

Adds a context menu to the browser extension so that users can enable the extension on domains without specifying the URL.

screenshot 2018-08-30 13 06 40

Testing plan:

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


chrome/extension/background.tsx
```diff
@@ -259,6 +263,32 @@ runtime.onMessage((message, _, cb) => {
...
+                try {
+                    const url = new URL(tab.url)
+                    permissions.request([url.origin])
+                } catch {
+                    console.error('Could not get tab url to request permissions')
+                }
+            },
+        },
+        () => {
+            storage.setSync({ hasEnableDomainContextMenu: true })
```
chrismwendt: I'd move this storage operation to the migration `createEnableDomainContextMenu.then(() => storage.setSync(...))` to keep the logic closer together

PR #126, opened 2018-09-04
chore: rename CXP -> Sourcegraph extensions

chrismwendt:


app/components/options/FeatureFlagCard.tsx
```diff
@@ -91,7 +91,7 @@ export class FeatureFlagCard extends React.Component<Props, {}> {
                                 <FormGroup check={true}>
                                     <Label check={true}>
                                         <Input onClick={this.onUseCXPToggled} defaultChecked={useCXP} type="checkbox" />{' '}
-                                        Enable CXP
+                                        Use Sourcegraph extensions
```
sqs: This `Enable CXP` is in the demo video at https://www.youtube.com/watch?v=j1eWBa3rWH8, unfortunately. Until @ryan-blunden can make a new video, I think you should make it `Use Sourcegraph extensions (enable CXP)` or something. Maybe you have a better idea.

chrismwendt: `Use Sourcegraph extensions (enable CXP)` sounds good since it emphasizes that these are called "Sourcegraph extensions" while still providing a bridge for users coming from the video.

PR #128, opened 2018-09-04
chore: remove some safari references

ijsnow:
Maintaining a Safari extension is no longer a priority for Sourcegraph. We are focusing on Chrome and Firefox as they have similar browser APIs and have the support of the majority of our user base.


README.md
```diff
@@ -8,7 +8,7 @@
 
 [![chrome version](https://img.shields.io/chrome-web-store/v/dgjhfomjieaadpoljlnidmbgkdffpack.svg?logo=Google%20Chrome&logoColor=white)](https://chrome.google.com/webstore/detail/sourcegraph/dgjhfomjieaadpoljlnidmbgkdffpack)
 [![chrome users](https://img.shields.io/chrome-web-store/users/dgjhfomjieaadpoljlnidmbgkdffpack.svg)](https://chrome.google.com/webstore/detail/sourcegraph/dgjhfomjieaadpoljlnidmbgkdffpack)
-[![chrome rating](https://img.shields.io/chrome-web-store/rating/dgjhfomjieaadpoljlnidmbgkdffpack.svg)](https://chrome.google.com/webstore/detail/sourcegraph/dgjhfomjieaadpoljlnidmbgkdffpack)  
+[![chrome rating](https://img.shields.io/chrome-web-store/rating/dgjhfomjieaadpoljlnidmbgkdffpack.svg)](https://chrome.google.com/webstore/detail/sourcegraph/dgjhfomjieaadpoljlnidmbgkdffpack)
```
felixfbecker: Please revert this change, this put the badges on the same line.

ijsnow: Prettier did that. I don't see anything that looks weird about the badges right now.

felixfbecker: If it did then CI would have failed on master - Prettier knows that spaces at the end of lines in markdown are significant, so it doesn't trim them. I just ran it locally and it doesn't from the CLI nor from VS Code. Must be something with your editor?

felixfbecker: I'm also fine with putting a `<br>` or `\` in there if that works and it's impossible to configure your editor to respect these

felixfbecker: It's also not editorconfig: https://github.com/sourcegraph/browser-extensions/blob/674cb8c55b6478bbfc46ad709f9848c0ab8cc4df/.editorconfig#L10-L11

PR #129, opened 2018-09-04
chore(deps): update rxjs

chrismwendt:
This is less hacky fix for npm/yarn link not working than https://github.com/sourcegraph/browser-extensions/pull/102


app/components/BlobAnnotator.tsx
```diff
@@ -355,7 +355,7 @@ export class BlobAnnotator extends React.Component<Props, State> {
                         )
                     })
                 )
-                .subscribe(data => {
+                .subscribe((data: TooltipData) => {
```
felixfbecker: Are you sure this is needed?

chrismwendt:
I forgot to re-run `npm run dev` after upgrading. There are still a bunch of type errors after linking. I'll have to figure this out later.

app/components/BlobAnnotator.tsx
chrismwendt: Without this annotation, TypeScript thinks `data` has type `{}`

felixfbecker: Then the type of the Observable is already wrong and should be fixed (not in the function)
This would probably even give a type error with `strictFunctionTypes`

sqs:
This is blocking me from updating browser-extensions' use of @sourcegraph/extensions-client-common and @sourcegraph/sourcegraph.proposed (nee `cxp`/`cxp-js`). I will look into this.

app/components/BlobAnnotator.tsx
sqs: Calling `pipe` 2x works :tada:

PR #132, opened 2018-09-04
feat: remove the Sourcegraph extension list

chrismwendt:
:warning: Don't merge this until the corresponding PR in sourcegraph/sourcegraph is merged (TODO(chris) add a link here)

This removes the extensions list from the browser extension. Adding/removing extensions will be handled on the Sourcegraph instance.

Resolves https://github.com/sourcegraph/browser-extensions/issues/127

TODO

- [ ] Add communication between the browser extension and the Sourcegraph instance for adding/removing/enabling/disabling extensions

PR #146, opened 2018-09-07
feat: accept settings updates from Sourcegraph pages

chrismwendt:
This connects the browser extension to Sourcegraph pages so that Sourcegraph can instruct the browser extension to edit client settings on its behalf.

This will allow us to eliminate the extensions list from the browser extension (see #132) and from e-c-c (see TODO).

I didn't piggyback on the existing `document.dispatchEvent` code because we might want to embed Sourcegraph in an iframe, in which case I think the origin will be different.


src/libs/sourcegraph/inject.tsx
```diff
@@ -18,3 +40,55 @@ function dispatchSourcegraphEvents(marker: HTMLElement): void {
...
+ * pings back. This always resolves to true.
+ */
+function connectToPage(): Promise<boolean> {
+    return new Promise(resolve => {
+        const ping = () => {
+            window.postMessage({ source: 'Client' as Source, type: 'Ping' as MessageType }, '*')
+        }
+
+        let isConnected = false
+        window.addEventListener('message', event => {
```
felixfbecker: This doesn't cleanup the listener after the Promise is resolved, so nothing here can be garbage collected. I would recommend just using `fromEvent`+`filter`+`toPromise`.

src/shared/backend/extensions.ts
```diff
@@ -319,3 +281,43 @@ export function createExtensionsContextController(
...
 }
+
+export const updateExtensionSettings = (
+    subjectID,
+    args: { extensionID: string; edit?: ConfigurationUpdateParams; enabled?: boolean; remove?: boolean }
+): Observable<undefined> => {
+    if (subjectID !== 'Client') {
+        return throwError('Cannot update settings for ' + subjectID + '.')
+    }
+    const update = new Subject<undefined>()
```
felixfbecker: Avoid Subjects when they are not needed. Here you can definitely get by with the Observable constructor. But actually a Promise is probably the better choice.

src/shared/backend/extensions.ts
```diff
@@ -319,3 +281,43 @@ export function createExtensionsContextController(
...
+    storage.getSync(storageItems => {
```
felixfbecker: Any errors within in this callback are not propagated to the caller. Wrap this call (and only this call) in a Promise or Observable (e.g. `bindCallback`), then do the rest of the logic in operators/`.then()`.

src/shared/components/options/ConnectionCard.tsx
```diff
@@ -74,12 +78,14 @@ export class ConnectionCard extends React.Component<Props, State> {
                 </div>
             )
         }
-        const hasPermissions = this.contentScriptUrls.every(val => permissionOrigins.indexOf(`${val}/*`) >= 0)
-        if (!hasPermissions && !permissionOrigins.includes('<all_urls>')) {
+        const forbiddenUrls = permissionOrigins.includes('<all_urls>')
+            ? []
+            : this.contentScriptUrls.filter(url => permissionOrigins.indexOf(`${url}/*`) === -1)
```
felixfbecker: `.includes()`

src/libs/sourcegraph/inject.tsx
```diff
@@ -1,20 +1,44 @@
...
+import storage from '../../browser/storage'
+import { updateExtensionSettings } from '../../shared/backend/extensions'
+
 export function injectSourcegraphApp(marker: HTMLElement): void {
     if (document.getElementById(marker.id)) {
         return
     }
 
+    // Generate and insert DOM element, in case this code executes first.
+    document.body.appendChild(marker)
```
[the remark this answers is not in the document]
felixfbecker: If it doesn't wait for DOM load, how do you ensure that `document.body` has already loaded? Afaik it's not safe to do DOM manipulation until the DOM is loaded

src/shared/backend/extensions.ts
```diff
@@ -315,3 +277,41 @@ export function createExtensionsContextController(
...
+                    clientSettings = applyEdits(
+                        clientSettings,
+                        typeof args.enabled === 'boolean'
+                            ? setProperty(clientSettings, ['extensions', args.extensionID], args.enabled, format)
+                            : removeProperty(clientSettings, ['extensions', args.extensionID], format)
+                    )
+                }
+                return new Promise<undefined>(resolve =>
+                    storage.setSync({ clientSettings }, () => {
+                        resolve(undefined)
```
felixfbecker: Can't you pass `resolve` directly? Since the callback has no parameters, it should fit the type `void`

src/libs/sourcegraph/inject.tsx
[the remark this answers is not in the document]
chrismwendt: This function only gets called when the DOM is loaded:

https://sourcegraph.sgdev.org/github.com/sourcegraph/browser-extensions@0c49ab1c4677356d6a680d65cd68bb81200eeecf/-/blob/src/extension/scripts/inject.tsx#L111-116

https://sourcegraph.sgdev.org/github.com/sourcegraph/browser-extensions/-/blob/src/extension/scripts/inject.tsx#L87:13&tab=references

src/shared/backend/extensions.ts
chrismwendt: I generally avoid that because it makes me less confident that the return value of the Promise will be what I intend for it to be.

src/libs/sourcegraph/inject.tsx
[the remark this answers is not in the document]
felixfbecker: Then I don't understand your above comment. Was the issue that this event handler would get dispatched before the one that adds the marker, and now both add the marker?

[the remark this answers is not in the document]
chrismwendt: Oh yeah, what I said was ambiguous. If you look at the red part of the diff where this statement used to be, you'll see that it's inside a callback to the `load` event, which executes asynchronously with respect to this function call.

src/shared/backend/extensions.ts
felixfbecker: Got it, if that's a concern I would eliminate it with in a different way with `.then(() => undefined)`. That keeps logic in Promise constructors minimal (no point for the next dev to think "I need to add this logic, I can just add it here before the `resolve` call")

Issue #154, opened 2018-09-10
Update to Babel 7


felixfbecker:
This brings a bunch of improvements, but most importantly in the webapp it made async/await "just work". Promises have a lot of caveats that frequently come up in code review so this would be great to remedy. We already did this in the web app, so this should be trivial to apply here. See https://github.com/sourcegraph/sourcegraph/pull/13006

sqs:
done in 4f84e33b8c278ca829b4e5564fe59b2f50f723e8

PR #165, opened 2018-09-13
build: support async functions, clean up babel usage

sqs:
This modernizes our Babel pipeline:

- Upgrades babel and related dependencies
- Removes unneeded transformations
- Emits JS supported by the last 3 versions of Chrome/Firefox (no need for older browser support in a browser extension, and for Phabricator our customers' browsers fit within this browserslist query)

I did this to fix an issue where `regeneratorRuntime` was required, because the `sourcegraph` npm package now uses async functions. I could have configured babel to properly use `regeneratorRuntime`, etc., but we don't need to babelify async functions anyway because this code executes only on browsers that support async functions natively. Making babel do less was easier (and sufficient for our needs) vs making babel do more and do it correctly.


ijsnow:
Nice, haven’t looked at the code, will do tomorrow. If you’re wanting to
merge before then, just please be sure to test in Firefox. We used to use a
sync functions but stopped because Babel poly fills were breaking Firefox.
It sounds like that won’t be a problem now if we’re just not transpiling
them now

felixfbecker:
Closes #154?

.babelrc
```diff
@@ -1,30 +1,11 @@
...
-    }
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "useBuiltIns": "entry",
+        "modules": false
+      }
+    ]
+  ]
```
felixfbecker: The new way for configuration is `babel.config.js`, check how we do it in the webapp (the config resolution is different so this is somewhat important)

package.json
```diff
@@ -23,7 +23,14 @@
       "@commitlint/config-conventional"
     ]
   },
+  "browserslist": [
+    "last 3 Chrome versions",
+    "last 3 Firefox versions"
```
felixfbecker: Why do we support the `last 1 version` in the webapp but the last 3 in the browser extension?

src/config/polyfill.tsx
```diff
@@ -22,3 +22,6 @@ require('intersection-observer')
 if (!GLOBAL.FormData.prototype.set) {
     require('formdata-polyfill')
 }
+
+// This gets automatically expanded into imports that only pick what we need.
+import '@babel/polyfill'
```
felixfbecker: This should be at the very top, because it's the most fundamental built-ins, so the other more specific polyfills are likely to depend on them

webpack/base.config.ts
```diff
@@ -41,11 +48,12 @@ export default {
             {
                 test: /\.tsx?$/,
                 use: [
-                    'babel-loader',
+                    babelLoader,
                     {
                         loader: 'ts-loader',
                         options: {
                             compilerOptions: {
+                                target: 'es6',
```
felixfbecker: This is the same as in tsconfig?

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "outDir": "./out",
-    "target": "es2015",
+    "target": "es6",
```
felixfbecker: es6 === es2015

sqs:
Cool - wanna just take over this PR?

package.json
felixfbecker: @sqs if you could reply here I can take over

sqs: No particular reason. The restart cycle took a long time to try various N values here so I just started at 3, it worked, and I didn’t try narrowing it.

ijsnow:
@sqs / @felixfbecker fixed the CI error. We still needed Uglify and it also needed to be updated to handle the new babel stuff

webpack/prod.config.ts
```diff
@@ -12,11 +12,13 @@ export default {
...
-                    // which causes Chrome to complain that the bundle is not UTF8
-                    ascii_only: true,
-                    beautify: false,
+                uglifyOptions: {
+                    output: {
+                        // Without this, Uglify will change \u0000 to \0 (NULL byte),
+                        // which causes Chrome to complain that the bundle is not UTF8
+                        ascii_only: true,
+                        beautify: false,
+                    },
```
felixfbecker: This is missing important options from the webapp https://github.com/sourcegraph/sourcegraph/blob/3a35f2ddaefad0feaffc557afbfe3c5eedd8ea0c/web/webpack.config.ts#L58

Issue #175, opened 2018-09-14
Duplicate hover contents


sqs:
Repro:

1. "Use Sourcegraph extensions" = on
1. Go to https://github.com/theupdateframework/notary/blob/fb795b0bc868746ed2efa2cd7109346bc7ddf0a4/server/storage/rethinkdb.go
1. Hover over a token

The hover contents are duplicated, as though it's fetching them via the builtin LSP client and via the `langserver/*` extensions and merging them.

sqs:
Yes, I removed it because it was broken (no Router context, needed for the Links). No extensions other than the langserver/go extension. I’ll dig into it more

chrismwendt:
I saw this briefly while working on something else, but I don't have a consistent repro.

chrismwendt:
I noticed that this bug only manifests while the options page is open. I suspect this is happening because https://github.com/sourcegraph/browser-extensions/blob/903acddcdd690339e491668852cbddb062c034c5/src/extension/scripts/background.tsx#L462 is being executed both in the background script and the options page (via a transitive `import`). That line it should only execute in the background script. To avoid this double execution, I'm looking into either splitting up the imports or conditionally executing that based on which script is running.

chrismwendt:
Submitted a fix https://github.com/sourcegraph/browser-extensions/pull/182. The root cause was that I was transitively importing background.tsx from the options page, which caused multiple port listeners to be added, which caused duplicate messages to be sent to the content scripts.

I plan to completely prevent transitive imports of the entrypoints in another PR.

PR #184, opened 2018-09-18
fix: prevent transitive importing of entrypoints

chrismwendt:
This prevents accidentally importing one entrypoint in another, which was the root cause of https://github.com/sourcegraph/browser-extensions/issues/175

This works by tagging each entrypoint in webpack with `window.EXTENSION_ENV = <name of entrypoint>`, having each entrypoint call `assertEnv(<name of entrypoint>)`, and throwing an error if `window.EXTENSION_ENV !== <name of entrypoint>`.

This also removes existing violations of these assertions :sparkles:


src/shared/components/options/BrowserSettingsEditor.tsx
```diff
@@ -1,10 +1,10 @@
 // We want to polyfill first.
```
ijsnow: Ah, was this the entry point for the extensions page? Just confused about the name change and don't remember this file before.

src/shared/components/options/ExtensionRegistry.tsx
```diff
@@ -7,10 +7,10 @@ import {
 import * as React from 'react'
 import { RouteComponentProps } from 'react-router-dom'
 import { Subscription } from 'rxjs'
-import { BrowserSettingsEditor } from '../../../extension/scripts/extensions'
 import { createExtensionsContextController } from '../../../shared/backend/extensions'
 import { GQL } from '../../../types/gqlschema'
 import { sourcegraphUrl } from '../../util/context'
+import { BrowserSettingsEditor } from './browserSettingsEditor'
```
ijsnow: Should the file name in this import be UpperCamelCase?

src/shared/messaging.ts
```diff
@@ -0,0 +1,18 @@
...
+    jsBundleURL: string
+}
+
+/**
+ * Executes the callback only on the first message that's received on the port.
+ */
+export const onFirstMessage = (port: chrome.runtime.Port, callback: (message: any) => void) => {
+    const cb = message => {
+        port.onMessage.removeListener(cb)
+        callback(message)
```
ijsnow: Nifty logic. Wouldn't have thought of that.

src/shared/components/options/BrowserSettingsEditor.tsx
chrismwendt: It might have been the entrypoint a while back, but currently the extensions page is part of the options page, so this entrypoint is unnecessary.

src/shared/components/options/ExtensionRegistry.tsx
chrismwendt: Yes, fixed.

PR #186, opened 2018-09-18
Code intel inject

ijsnow:
This PR changes GitHub to use the new generic way of injecting code intelligence. 

A couple TODOs:

- [x] Add single file view, snippets and search results back, won't take long
- ~Talk with @chrismwendt about what I blew away in terms of Extensions and how to get it added back~
- [x] Put behind a feature flag for GitHub until we get extensions support back in

Testing plan:

Go through the regular flows of using 

I have tested on:

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


src/libs/code_intelligence/code_views.ts
```diff
@@ -0,0 +1,128 @@
...
+    const intersectionObserver = new IntersectionObserver(
+        entries => {
+            for (const entry of entries) {
+                // `entry` is an `IntersectionObserverEntry`,
+                // which has
+                // [isIntersecting](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/isIntersecting#Browser_compatibility)
+                // as a prop, but TS complains that it does not
+                // exist.
+                if ((entry as any).isIntersecting) {
+                    intersectingElements.next(entry.target as HTMLElement)
```
felixfbecker: Instead of Subjects, prefer the Observable constructor

src/libs/code_intelligence/code_views.ts
```diff
@@ -0,0 +1,128 @@
...
+        {
+            rootMargin: `${margin}px`,
+            threshold: 0,
+        }
+    )
+
+    return (codeViews: Observable<ResolvedCodeView>) =>
+        new Observable<ResolvedCodeView>(observer => {
+            codeViews.subscribe(({ codeView, ...rest }) => {
+                intersectionObserver.observe(codeView)
```
felixfbecker: It seems wrong to have the intersectionObserver be shared between all Observable subscriptions at this layer - if the intend is to share, the `share()` operator should accomplish that

src/libs/code_intelligence/code_views.ts
```diff
@@ -0,0 +1,128 @@
...
+        mergeMap(container =>
+            from(codeHost.codeViews!).pipe(
+                map(({ selector, ...info }) => ({
+                    info,
+                    matches: container.querySelectorAll<HTMLElement>(selector),
+                }))
+            )
+        ),
+        mergeMap(({ info, matches }) =>
+            of(...matches).pipe(
```
felixfbecker: `from` creates Observables from arrays

src/libs/code_intelligence/code_views.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    if (watchChildrenModifications) {
+        const possibleLazilyLoadedContainers = new Subject<HTMLElement>()
+
+        const mutationObserver = new MutationObserver(mutations => {
+            for (const mutation of mutations) {
+                if (mutation.type === 'childList' && naiveCheckIsHTMLElement(mutation.target)) {
+                    const { target } = mutation
+
+                    possibleLazilyLoadedContainers.next(target)
```
felixfbecker: Prefer Observable constructor over Subject

src/libs/github/code_intelligence.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+const diffCodeView: CodeViewWithOutSelector = {
+    dom: diffDomFunctions,
+    getToolbarMount: createCodeViewToolbarMount,
+    resolveFileInfo: resolveDiffFileInfo,
+    toolbarButtonProps,
+}
+
+const resolveCodeView = (elem: HTMLElement) => {
```
felixfbecker: Add return type

src/libs/github/code_intelligence.ts
```diff
@@ -0,0 +1,45 @@
+import { CodeHost, CodeViewResolver, CodeViewWithOutSelector } from '../code_intelligence'
+import { diffDomFunctions } from './dom_functions'
+import { resolveDiffFileInfo } from './file_info'
+import { createCodeViewToolbarMount, parseURL } from './util'
+
+const toolbarButtonProps = {
```
felixfbecker: Add type declaration

src/libs/github/file_info.ts
```diff
@@ -0,0 +1,64 @@
...
+    of(codeView).pipe(
+        map(codeView => {
+            const { repoPath } = parseURL()
+
+            return { codeView, repoPath }
+        }),
+        map(({ codeView, ...rest }) => {
+            const { headFilePath, baseFilePath } = getDeltaFileName(codeView)
+            if (!headFilePath) {
+                throw new Error('cannot determine file path')
```
felixfbecker: According to the type this never happens, in what case does it happen?

src/libs/github/file_info.ts
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+            return { ...rest, codeView, headFilePath, baseFilePath }
+        }),
+        map(({ codeView, ...rest }) => {
+            const diffResolvedRev = getDiffResolvedRev()
+            if (!diffResolvedRev) {
+                throw new Error('cannot determine delta info')
```
felixfbecker: In what case does this happen? This error message is rather unhelpful, should `getDiffResolvedRev` maybe throw the error instead?

src/libs/phabricator/code_intelligence.ts
```diff
@@ -126,3 +127,19 @@ export const phabCodeViews: CodeView[] = [
         toolbarButtonProps,
     },
 ]
+
+function checkIsPhabricator(): Promise<boolean> | boolean {
```
felixfbecker: Could this just always return a Promise for simplicity? You could just make it an async function

src/libs/code_intelligence/code_intelligence.tsx
```diff
@@ -196,61 +224,88 @@ function initCodeIntelligence(codeHost: CodeHost): { hoverifier: Hoverifier } {
...
+                        mount
+                    )
                 )
             })
     )
+
+    return subscriptions
+}
+
+async function injectCodeIntelligenceToCodeHosts(codeHosts: CodeHost[]): Promise<Subscription> {
```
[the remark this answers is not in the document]
ijsnow: Should I not make this an `async` function?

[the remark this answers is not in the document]
felixfbecker: It can be (it looks like it does only side effects)
I just thought the returned Subscription was weird because if it is to cancel the operation, the operation already happened when the Promise resolves. But I see that it actually represents the subscription for hoverify etc? In that case this makes sense. Just some documentation would help explain what the return value actually is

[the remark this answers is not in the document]
ijsnow: Yea, I'll add some docs

src/libs/code_intelligence/code_views.ts
```diff
@@ -0,0 +1,128 @@
+import { from, merge, Observable, of, Subject } from 'rxjs'
+import { filter, map, mergeMap } from 'rxjs/operators'
+
+import { CodeHost, ResolvedCodeView } from './code_intelligence'
+
+/**
+ * Cast a Node to an HTMLElement if it has a classList. This should not be used
+ * if you need 100% confidence the Node is an HTMLElement.
+ */
+function naiveCheckIsHTMLElement(node: Node): node is HTMLElement {
```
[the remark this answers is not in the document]
ijsnow: Well, tbh didn't really think about it. I get why `Node`, `Element`, and `HTMLElement` all exists as separate types but they do confuse me sometimes.

src/libs/code_intelligence/code_views.ts
ijsnow: To be honest, I'm not sure how to use that in place of this subscription here. That sounds reasonable. Could you give me some guidance?

src/libs/code_intelligence/code_views.ts
ijsnow: This is probably related to your comment below but I found it quite hard to implement this in another way. I only ever want to create on `IntersectionObserver` so I put it above the function that gets called for everything that's piped through but it needs to be used there so I created a `Subject` so data could be piped into it.

ijsnow: I tried using the observable constructor but ended up here

src/libs/github/file_info.ts
ijsnow: There isn't an error as far as I know. This piece was just ported over from existing logic. Would probably be worth looking into at some point.

ijsnow:
This is now behind a feature flag. We don't need to worry about blowing away extensions stuff before merging this in. cc @chrismwendt

PR #192, opened 2018-09-21
build: install cypress binary

ijsnow:
The build is failing for release because the binary isn't installed.


.travis.yml
```diff
@@ -41,8 +41,6 @@ jobs:
       script:
         - yarn run test:e2e
     - stage: release
-      env:
-        - CYPRESS_INSTALL_BINARY=0
```
felixfbecker: I don't understand why Cypress is needed for releasing

PR #197, opened 2018-09-26
feat: Gitlab

ijsnow:
This PR brings support for GitLab.

## Testing plan:

Run the browser extension and visit the following and check to ensure it works

### Code intelligence

* [Diff view with different versions](https://gitlab.com/pantomath-io/panto/merge_requests/348/diffs?view=parallel)
* [Changed file name](https://gitlab.com/ijsnow/hello-gitlab/merge_requests/1/diffs)
* [Diff that’s good for testing misalignment](https://gitlab.com/pantomath-io/panto/merge_requests/348/diffs)
* [Commit page](https://gitlab.com/pantomath-io/panto/commit/3b1dd7ca2039a925e9dba25b402240e1ab98310f)

### Search

*Action*: Enable search feature flag and [open this page](https://gitlab.com/search?utf8=%E2%9C%93&search=state&group_id=&project_id=3043340&search_code=true&repository_ref=). 

*Expected result*: A new tab should open on the sourcegraph search results page with the `repo:filter` and search query in the search bar

I have tested on:

- [x] Chrome
- [x] Firefox


src/libs/code_intelligence/search.ts
```diff
@@ -0,0 +1,79 @@
...
+    repoPath,
+    rev,
+    query,
+}: SearchPageInformation): { url: string; repo: string; rev: string | undefined; query: string } | undefined {
+    if (repoPath) {
+        if (rev) {
+            return {
+                url: `search?q=${encodeURIComponent(query)}&sq=repo:%5E${encodeURIComponent(
+                    repoPath.replace(/\./g, '\\.')
+                )}%24@${encodeURIComponent(rev)}&utm_source=${getPlatformName()}`,
```
felixfbecker: I would use the `URL` API to avoid manual `encodeURIComponent` calls

src/libs/code_intelligence/search.ts
```diff
@@ -0,0 +1,79 @@
...
+            query: `repo:^${repoPath.replace(/\./g, '\\.')}$ ${query}`,
+        }
+    }
+}
+
+export function initSearch({ getRepoInformation, checkIsSearchPage }: SearchFeature): void {
+    if (checkIsSearchPage()) {
+        storage.getSync(({ executeSearchEnabled }) => {
+            // GitHub search page pathname is <org>/<repo>/search
+            if (false && !executeSearchEnabled) {
```
felixfbecker: ?

src/libs/gitlab/api.ts
```diff
@@ -0,0 +1,71 @@
...
+interface DiffVersionsResponse {
+    base_commit_sha: string
+}
+
+type GetBaseCommitIDInput = Pick<GitLabDiffInfo, 'owner' | 'repoName' | 'mergeRequestID' | 'diffID'>
+
+const buildURL = (owner: string, repoName: string, path: string) =>
+    `${window.location.origin}/api/v4/projects/${owner}%2f${repoName}${path}`
+
+const get = <T>(url: string): Observable<T> => ajax.get(url).pipe(map(({ response }) => response as T))
```
felixfbecker: Avoid type parameters that are only used in the return type - it's only a cast in disguise. It's better to cast explicitly.

src/libs/gitlab/scrape.ts
```diff
@@ -0,0 +1,172 @@
...
+export function getPageInfo(): GitLabInfo {
+    const host = window.location.hostname
+
+    const parts = window.location.pathname.slice(1).split('/')
+
+    const owner = parts[0]
+    const repoName = parts[1]
+
+    let pageKind: GitLabPageKind
+    if (window.location.pathname.match(new RegExp(`${owner}/${repoName}/commit`))) {
```
felixfbecker: Wouldn't `pathname.includes()` do the same? Do you have to escape characters in the regex?

src/libs/gitlab/scrape.ts
```diff
@@ -0,0 +1,172 @@
...
+    const filePath = parts.slice(4).join('/')
+
+    return {
+        repoPath,
+        filePath,
+        rev,
+    }
+}
+
+const createErrorBuilder = (message: string) => (kind: string) => new Error(`${message} (${kind})`)
```
felixfbecker: What is the purpose of `kind`? It looks more like an error code than meant for human consumption, but it is put into the `message`.

src/libs/code_intelligence/search.ts
ijsnow: Lol good catch. Was too lazy to switch the feature flag while working

src/libs/gitlab/api.ts
ijsnow: Casting is actually what I intended with that. I think it's easier to read at the call sights.

src/libs/gitlab/scrape.ts
ijsnow: It's added for making it easier to find the site of the error. Maybe `kind` is just a bad name.

src/libs/gitlab/api.ts
```diff
@@ -0,0 +1,71 @@
...
+    base_sha: string
+    head_sha: string
+    start_sha: string
+}
+
+/**
+ * Response from the GitLab API for fetching a merge request. Note that there
+ * is more information returned but we are not using it.
+ */
+interface MRResponse {
```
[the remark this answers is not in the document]
ijsnow: That's how I originally had it and thought `ResuestResponse` sounded weird. This comment was enough to convince me to ignore that and spell it out

PR #199, opened 2018-09-27
feat: support Sourcegraph extensions in the new inject code

chrismwendt:
- Adds support for Sourcegraph extensions with the `newInject` feature flag
- Adds support for hovers/defs for Sourcegraph extensions on diff pages
- Bumps e-c-c and sourcegraph-extension-api


src/libs/code_intelligence/extensions.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export function injectExtensionsGlobalComponents({
+    extensionsController,
+    extensionsContextController,
+}: {
+    extensionsController: ClientController<ConfigurationSubject, Settings>
+    extensionsContextController: Controller<ConfigurationSubject, Settings>
+}): void {
+    const headerElem = document.querySelector('div.HeaderMenu>div:last-child')
```
ijsnow: @chrismwendt This isn't abstract. Nothing inside `libs/code_intelligence` should know *anything code host specific*. That's why I had created a "getCommandPaletteMount` function that the code host has to implement.

src/libs/code_intelligence/code_intelligence.tsx
```diff
@@ -131,13 +157,53 @@ export interface FileInfo {
     baseHasFileContents?: boolean
 }
 
+interface Controllers {
+    ctx: Controller<ConfigurationSubject, Settings>
+    ctl: ClientController<ConfigurationSubject, Settings>
```
ijsnow: Could we have a little more specific names/doc blocks? Since the interface is named `Controllers`, `ctx` and `ctl` just confuses me especially considering the types look like they indeed are both controllers

src/libs/code_intelligence/extensions.tsx
chrismwendt: Oh right, that was unintentional and definitely should be factored out.

ijsnow:
TODO: Implement `CodeView.getLineRanges` for GitHub and Phabricator, or make it optional behind an extensions field that is optional.

chrismwendt:
Fixed the extensions list page and pushed.

PR #245, opened 2018-10-14
feat: use access tokens for authentication

ijsnow:
This PR changes request logic to use access tokens. This means that admins won't have to add their code hosts to the `corsOrigin` anymore.

This fixes `corsOrigin` errors because we are able to make requests to the API from the options menu. Any request from the options menu will create an access token. I landed on creating a token on each request rather than explicitly creating a token when entering and saving a new URL because multiple requests go out at different times and it was difficult to make this happen before other requests go out. The easiest way was to get to the root of each request. 

Without this architecture change/decision, some requests would fail with a 401 which could cause problems. You would have to click save and then refresh the page for everything to work properly.

@felixfbecker this is an rxjs-y one so I'd appreciate a review from that perspective.

Testing plan:

Spin up a sourcegraph instance and put a private repo on there. Ensure that `corsOrigin` is not set in the site config. Point your browser extension at http://localhost:3080 or wherever the instance you just spun up is hosted at and click save. Go to http://localhost:3080/users/isaac/account/tokens and ensure there is a token there for your browser extension (chrome/firefox). Ensure code intelligence works on the private repository you added.

I have tested on:

- [x] Chrome
- [x] Firefox
- [x] Phabricator Bundle <- I'll test this before merging in because there are potential code paths that could break this. EDIT: Tested.


src/shared/auth/access_token.ts
```diff
@@ -0,0 +1,37 @@
+import { omit } from 'lodash'
+import { Observable } from 'rxjs'
+import { switchMap } from 'rxjs/operators'
+import storage from '../../browser/storage'
+
+export const getAccessToken = (url: string): Observable<string | undefined> =>
+    new Observable(observer => {
+        storage.getSync(items => {
+            observer.next(items.accessTokens[url])
```
chrismwendt: Careful, what if `accessTokens` is not defined?

src/shared/backend/graphql.tsx
```diff
@@ -39,44 +47,60 @@ function requestGraphQL(
...
-                    throw authError
-                }
-                delete repoUrlCache[ctx.repoKey]
-                // We just tried the last url
-                throw err
-            }
-            return requestGraphQL(ctx, request, variables, DEFAULT_SOURCEGRAPH_URL, retry, authError)
-        })
+                        if (headers && headers.authorization) {
+                            // If we got a 401 with a token, get rid of the and
```
chrismwendt: Typo somewhere

src/shared/backend/auth.ts
```diff
@@ -0,0 +1,54 @@
...
+                false
+            )
+        ),
+        map(({ data, errors }) => {
+            if (!data || !data.createAccessToken || (errors && errors.length > 0)) {
+                throw createAggregateError(errors)
+            }
+            return data.createAccessToken.token
+        }),
+        share()
```
chrismwendt: Can you explain what `share()` does in this context? I skimmed over an article and docs about it, but I'm still not sure how/why it's useful.

chrismwendt:
Oh, I just realized that the access token is created from the options page, which doesn't require `corsOrigin`.

src/shared/backend/headers.tsx
```diff
@@ -1,40 +1,93 @@
...
+            //
+            // - An Origin header with a URI that Sourcegraph trusts. The prod and dev Chrome extension origins
+            //   (chrome-extension://...) are trusted, and site admins can specify other trusted origins in the site config
+            //   "corsOrigin" property.
+            // - An X-Requested-With header (with any nonempty value). This tells the server that the request is from a
+            //   trusted origin, because that header could not be added to a cross-domain request unless it already passed
+            //   the server's CORS rules. (See
+            //   https://stackoverflow.com/questions/17478731/whats-the-point-of-the-x-requested-with-header for more
+            //   info.)
+            // - Using application/json as the Content-Type or Accept result in CORS blocking the request.
```
chrismwendt: Is this third bullet point intended to be one of the things required for the Sourcegraph server to accept the request? I don't see how it relates. BTW, this whole comment is nice :+1:

src/shared/backend/headers.tsx
```diff
@@ -1,40 +1,93 @@
...
+            // - Chrome (usually) automatically includes "Origin: chrome-extension://..." but does NOT allow us to include
+            //   other headers (such as X-Requested-With), or else they are blocked by the CSP policy of GitHub (or other
+            //   target page). Chrome sometimes sends "Origin: <window.location.origin>" instead; it's not clear in what
+            //   cases or why this occurs.
+            // - Safari includes "Origin: <window.location.origin>" (where "<window.location.origin>" is the
+            //   `window.location.origin` of the current page).
+            // - Firefox does NOT include any Origin header, so we need to send an "X-Requested-With" header.
+            const needsCORSHeader = getPlatformName() === 'firefox-extension' || isPhabricator
+            if (needsCORSHeader) {
+                headers.append('X-Requested-With', `Sourcegraph - ${getPlatformName()} v${getExtensionVersionSync()}`)
```
chrismwendt: Should this be `needsXRequestedWithHeader`?

src/shared/backend/auth.ts
```diff
@@ -0,0 +1,54 @@
...
+    )
+
+    return (userID: GQL.ID): Observable<string> => {
+        userIDs.next(userID)
+
+        return accessTokenCreator
+    }
+}
```
chrismwendt: Maybe I need to first understand what `share()` does, but I have a lot of questions about `buildAccessTokenCreator`:

- Would a pattern like storing a map from user ID to `Promise<Token>` work in this case?
- How does this actually ensure that two requests for an access token for the same user only result in one access token being created?
- How does this handle the case where two requests for access tokens for two different users?

I think some of my confusion is caused by not knowing which `Observable`s are expected to emit once or more than once, which would be solved by using `Promise`s, but I understand those can't be canceled.

src/shared/backend/headers.tsx
```diff
@@ -1,40 +1,93 @@
...
+            }
+
+            return headers
+        }),
+        // rxjs ajax seems to not like the Header type so we should reconstruct it as a plain object.
```
felixfbecker: Then why do we use `Headers` in the first place?

src/shared/backend/lsp.tsx
```diff
@@ -44,6 +44,21 @@ export function isEmptyHover(hover: HoverMerged | null): boolean {
     return !hover || !hover.contents || (Array.isArray(hover.contents) && hover.contents.length === 0)
 }
 
+const createRequest = (url: string, path: string, requests: any[]) =>
```
felixfbecker: It would be helpful to have a return type here (any named function really)

felixfbecker: `createRequest` to me sounds like a factory function for a `Request` object. I think an action verb like `request()` would make more sense, since none of our API request functions use a `create` prefix

src/shared/backend/lsp.tsx
```diff
@@ -52,15 +67,8 @@ export function sendLSPHTTPRequests(requests: any[], url: string = sourcegraphUr
...
         const urlPathHint = requests[1] && requests[1].method
-        return ajax({
-            method: 'POST',
-            url: `${urlToTry}/.api/xlang/${urlPathHint || ''}`,
-            headers: getHeaders(),
-            crossDomain: true,
-            withCredentials: true,
-            body: JSON.stringify(requests),
-            async: true,
-        }).pipe(
```
felixfbecker: Nice cleanup

src/shared/backend/lsp.tsx
felixfbecker: I find the `path` parameter confusing, as this always gets passed the `method` and is only put to the path for debug info. At first I thought below it was a bug because the method was passed to a parameter named `path`

src/shared/components/options/ConnectionCard.tsx
```diff
@@ -225,13 +232,15 @@ export class ConnectionCard extends React.Component<Props, State> {
...
-            () => {
-                this.setState(() => ({ site: undefined }))
-            }
+        this.subscriptions.add(
+            fetchSite().subscribe(
+                site => {
+                    this.setState({ site })
+                },
+                () => {
+                    this.setState({ site: undefined })
```
felixfbecker: I think it would be better if `site` had three states, `undefined` (loading), Error (unable to connect) and `ISite` (connected). Currently "Unable to connect" always flashes when loading the page because `undefined` is both error and loading state

src/shared/backend/headers.tsx
```diff
@@ -1,40 +1,93 @@
+import { Observable, of } from 'rxjs'
+import { filter, map, switchMap } from 'rxjs/operators'
+
+import { isDefined } from '@sourcegraph/codeintellify/lib/helpers'
```
felixfbecker: That's an internal module which should not be imported. Only the index file is public API, everything else may break in minor/patch versions. The utility is a one-liner so is not worth code-sharing and can be copied.

src/shared/backend/headers.tsx
ijsnow: @sqs wrote the comment (its on the left side of the diff too)

src/shared/backend/headers.tsx
ijsnow: Could be, just kept the old variable name(its on the left side of the diff).

src/shared/backend/headers.tsx
ijsnow: I was having trouble getting this to work without using it. I would prefer to use it anyways. There has to be a reason why `Header` is in the standard.

src/shared/backend/lsp.tsx
ijsnow: I'll change it to `method` but I don't find it that confusing because it becomes part of the path.

src/shared/components/options/ConnectionCard.tsx
ijsnow: This is existing logic(left side of the diff) and this component will be removed soon. We are re-doing the options menu soon.

src/shared/backend/auth.ts
ijsnow: No, the goal is not to just memoize the request. It's already saved in `browser.storage.sync` which can be used across page loads/tabs/even devices if they are signed into their browser. `memoizeObservable` just keeps an object in memory.

The goal of this is to only ever send out one request to create an access token at a time. `memoizeObservable` would be fine if we never had concurrent requests going out as we do on options page load.

Below is a diagram of what happens with only using `memoizeObservable` for concurrent requests(or if we were to use promises without some other similar and much more complicated implementation of the same thin - more complicated because rxjs is much easier to treat as a sharable pipeline).

screen shot 2018-10-15 at 11 07 11 am

What `buildAccessTokenCreator` does is create a single source observable that each concurrent request can use to get a single access token from. Below is a diagram of how it looks with concurrent requests.

screen shot 2018-10-15 at 11 14 24 am

src/shared/auth/access_token.ts
ijsnow: Added to migration. It'll be defined.

src/shared/auth/access_token.ts
```diff
@@ -0,0 +1,37 @@
...
+        switchMap(
+            token =>
+                new Observable<string>(observer => {
+                    storage.getSync(({ accessTokens }) =>
+                        storage.setSync({ accessTokens: { ...accessTokens, [url]: token } }, () => {
+                            observer.next(token)
+                            observer.complete()
+                        })
+                    )
+                })
```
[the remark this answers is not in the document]
ijsnow: Is the await at the bottom supposed to be the return?

src/shared/backend/auth.ts
ijsnow: I found [this example](https://www.learnrxjs.io/operators/multicasting/share.html) to be helpful for understanding the effect of `share()`. It essentially shares the side effects (access token creation) from each step in the pipeline before this share. Without it, even with `switchMap` cancellation, there would be two access tokens created if two requests were piped through here concurrently.

src/shared/backend/auth.ts
ijsnow: @chrismwendt 

> Would a pattern like storing a map from user ID to Promise<Token> work in this case?

It could work but would have a lot more complicated code and lose some benifits of rxjs. See my above comment.

> How does this actually ensure that two requests for an access token for the same user only result in one access token being created?

`switchMap` cancels old requests from previous inputs and `share` shares the single result with observer.

> How does this handle the case where two requests for access tokens for two different users?

It doesn't. Could you explain a case where we would need to handle that?

src/shared/auth/access_token.ts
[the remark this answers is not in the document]
ijsnow: I tried and the resulting code feels weird to me. It mixes patterns in one function declaration.

ijsnow:
@felixfbecker 

> Some high-level explanation in the PR description or commit message what this code change does would be helpful.
Here's my understanding from reading the code (which is an inefficient and error-prone way to distill this information): On every Sourcegraph API request, if we don't have an access token, we make an API request to the Sourcegraph API to get an access token and use that. But I feel like I'm missing something? How does exactly does this solve the CORS problem?

Your high level understanding is correct. This gets around the corsOrigin problem because the first requests to a Sourcegraph instance from the browser extension will always be from the options menu which is able to get around CORS via the same bypass logic we used to have for all requests coming from the extension.

I've updated the PR description to explain that.

src/shared/backend/auth.ts
felixfbecker: I still don't understand why `memoizeObservable` doesn't work.

> The goal of this is to only ever send out one request to create an access token at a time. memoizeObservable would be fine if we never had concurrent requests going out as we do on options page load.

That's exactly what `memoizeObservable` does. If a request is made, and then another is made (even before the first one returned), the second one will get the result of the first (once it's there). Where does this behavior deviate from what you want in this situation?

> > How does this handle the case where two requests for access tokens for two different users?
> It doesn't. Could you explain a case where we would need to handle that?

I was confused by this too, because even if the application can never get into that state, reading this function suggests that the returned function could be called multiple times with different values for `userID`, which would result in unexpected behavior.

ijsnow: I didn't realize that there was logic for that in `memoizeObservable` but I can tell you that it wasn't working without this.

ijsnow: > I was confused by this too, because even if the application can never get into that state, reading this function suggests that the returned function could be called multiple times with different values for userID, which would result in unexpected behavior.

Would renaming it to be `source` reduce your confusion?

felixfbecker:
Thanks for the explanation. Can it ever happen that the request is _not_ the first one?

ijsnow:
Nope, its always the first. Clicking save kicks off a request which will handle it.

src/shared/backend/auth.ts
ijsnow: Sorry for the misunderstanding @felixfbecker. I just tried again and this works the same as `memoizeObservable`. I must have not thought to use it before because theres nothing in the definition of the term `memoize` that says `memoizeObservable` will wait for other requests for the same thing and there's nothing in the comments about it. I'll push a new commit using it.

I read your initial comment on this above as "oh he hasn't heard about the `memoizeObservable` function yet so I'll just tell him to use it instead" which wasn't very helpful as its used all over all Sourcegraph code bases and didn't make me feel very good. When giving reviews, I like to assume there was a reason for everything.

ijsnow: It probably would work although it still wouldn't fit into the code around it very nicely. 

Anyway, I don't have a builder anymore. Well, I do but its `memoizeObservable`. And it takes a `userID` because its [a parameter in the graphql mutation](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/graphqlbackend/schema.graphql#L149).

src/shared/backend/server.ts
```diff
@@ -43,7 +44,10 @@ export const fetchCurrentUser = (): Observable<GQL.IUser | undefined> =>
                 }
                 siteAdmin
             }
-        }`
+        }`,
+        undefined,
+        undefined,
```
felixfbecker: This should use an options object

src/shared/auth/access_token.ts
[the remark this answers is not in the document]
felixfbecker: I don't care strongly about this, but I'd say it uses the right pattern in the right context. It's bad to mix equivalent patterns like callbacks and Promises, but Observables and Promises are different patterns (note that the current state mixes the callback pattern with Observables). Observables represent 0..n values, Promises a single value (or an operation / side effect sequence). It is therefor totally okay to e.g. map each single value of the 0..n values of an Observable to a single other value represented by a Promise. That's why `switchMap` etc all support returning a Promise/taking an async function.
Using the Observable constructor on the other hand is very uncommon inside a `switchMap` callback, it is usually only used for functions when needed. Usually you'd use an operator like `bindCallback` (which would also remove the mixing of callbacks and Observables and have the same protection benefit I pointed out in my previous comment).
The goal of the function here is more causing side effects than returning a stream of values, which Promises are more suitable for than Observables.

You are free to decide, just wanted to give some perspective 🙌

src/shared/backend/headers.tsx
felixfbecker: Yeah looks like `Headers` actually has some magic around CORS in it. We might want to consider just moving away from rxjs `ajax` and use the `fetch` API with `abortable-rx`

Issue #234, opened 2018-10-05
Codecov: Set API token only appears in public repo or when signed out


joebowbeer:
I'm noticing some weird behavior with what I think is called the context menu, which may be preventing me from viewing codecov for a private GitHub repo.

If I'm viewing a private repo, the context menu either doesn't display or it shows only the Codecov documentation item. Then when I view a public repo, the menu shows more items, including the Codecov Set API item.

Even at sourcegraph.com, if I'm signed in, then the context menu only shows me one item: Codecov: Documentation and support. Then, if I sign out, the context menu shows me seven items, including Codecov: Set API token. 

(By the way, I'm also confused about the connection between my browser extension settings and the ones for my account at sourcegraph.com. What is the connection between my sourcegraph account and my GitHub or Codecov account? This may be related to discussion at #232?)

sqs:
@chrismwendt 	Any idea why this is happening?

chrismwendt:
According to [the Codecov Sourcegraph extension's `package.json`](https://github.com/sourcegraph/sourcegraph-codecov/blob/2de3605a298422836982c56745cd371db144ad1a/package.json#L124-L156), the "Set API token for private repositories" action will only show up when you're viewing a file. That doesn't explain why when you sign out you see more items in the command palette (this is mysterious).

I double checked that I see Codecov line BG colors on private code, so this should work for you too.

@joebowbeer Could you share repro steps or a video on your Sourcegraph instance, and make sure that the Codecov Sourcegraph extension is enabled when you go to sourcegraph.example.com/extensions ?

joebowbeer:
Hi Chris, sorry for the late response

I'm not running a sourcegraph instance. I was under the impression that
sourcegraph.com could be made to work.

I did verify that the codecov extension is enabled (at
sourcegraph.com/extensions), but navigating to a source view in GitHub
still presents me with "*Configure Sourcegraph" at the top and a context
menu on the right (4 horizontal lines) that does not expand. The up/down
arrow simple changes direction. I assume the menu is empty.

I'm not sure what to expect anymore.

However, I think I may need to run a sourcegraph instance for other
reasons, and I'm guessing your extension might work better in that case.

Below are the questions I'm trying to answer now. I would really appreciate
knowing if you think I'm on the right track.

Background: we're publishing codecov to codecov.io and we want to make
codecov + private source visible to members of our GitHub organization.
(Search and intelligence and all that may be an added benefit in the
future.)

[...]

PR #249, opened 2018-10-17
feat: support running extensions on private code without a private Sourcegraph instance

chrismwendt:
**Motivation** This adds support for running extensions (e.g. Codecov) on private code even when the repository does not exist on the Sourcegraph instance. Doing so will resolve this issue: https://github.com/sourcegraph/browser-extensions/issues/234#issuecomment-430429015

**Problem** The browser extension must not let arbitrary GraphQL requests through to the Sourcegraph instance (which could be Sourcegraph.com) because they might contain private info (such as repo names), but some GraphQL requests are necessary to run extensions at all (e.g. fetching the extension manifests).

**Implementation** I've marked some GraphQL requests as safe by passing in `requestMightContainPrivateInfo: false` (defaults to `true` to keep old behavior for unmarked requests). The browser extension will allow those requests to be sent to the Sourcegraph instance. I also blocked `queryLSP` when sent from a private repo to Sourcegraph.com.

**Reviewers** The main file to review is `src/shared/backend/extensions.ts`. If you want to review the rest, I'd recommend reviewing commit-by-commit.

- @ijsnow I'd like your :eyes: on "refactor: avoid a network request while resolving the revision"
- @sqs I'd like your input on the interaction between e-c-c and the browser extension in "feat: allow GraphQL requests that are known to not contain private info" and "feat: block queryLSP from a private repo to Sourcegraph.com" and whether or not this approach makes sense

**Testing**

- `git checkout allow-known-safe-gql-requests`
- `yarn ; yarn run dev`
- Make sure one language extension (e.g. Go) and one other extension (e.g. token-highlights) have the following behavior in these scenarios:
  - Private repository that doesn't exist on a private Sourcegraph instance: token-highlights should work, Go should not (also, Go should not spam you with 404s in the hover tooltips)
  - Private repository that exists on a private Sourcegraph instance: both should work
  - Private repository that doesn't exist on Sourcegraph.com: token-highlights should work, Go should not (also, Go should not send ANY requests to Sourcegraph.com - you must check the background script's Network tab to confirm)
- Public repos should work as usual (check both the content script and background script for errors you haven't seen before)

Depends on https://github.com/sourcegraph/sourcegraph-langserver-http/pull/8
Depends on https://github.com/sourcegraph/extensions-client-common/pull/57


src/shared/util/dom.tsx
```diff
@@ -12,3 +12,31 @@ export function insertAfter(el: HTMLElement, referenceNode: Node): void {
 export function isMouseEventWithModifierKey(e: MouseEvent): boolean {
     return e.altKey || e.shiftKey || e.ctrlKey || e.metaKey || e.which === 2
 }
+
+export function commitIDFromPermalink({
```
ijsnow: *No action required for this PR*

This looks like it could be abstracted slightly more and used for other code hosts if we add a parameter `container` which defaults to `document`. In other code hosts, the commit is taken per code view, not per page(though most of the time each code view will be at the same commit for any page you're looking at).

src/shared/util/dom.tsx
```diff
@@ -12,3 +12,31 @@ export function insertAfter(el: HTMLElement, referenceNode: Node): void {
...
+    selector,
+    hrefRegex,
+}: {
+    selector: string
+    hrefRegex: RegExp
+}): string | Error {
+    const permalinkElement = document.querySelector(selector)
```
ijsnow: Based on this functons name, I would consider casting this to a link and then below we don't have to use `getAttribute`. 

```suggestion
    const permalinkElement = document.querySelector<HTMLAnchorElement>(selector)
```

src/shared/util/dom.tsx
```diff
@@ -12,3 +12,31 @@ export function insertAfter(el: HTMLElement, referenceNode: Node): void {
...
+    if (!permalinkElement) {
+        return new Error(
+            `Unable to determine the commit ID (40 character hash) you're on because the permalink shortcut element (query selector ${selector}) which contains the commit ID does not exist in the DOM.`
+        )
+    }
+    const href = permalinkElement.getAttribute('href')
```
ijsnow: If you accept the above suggestions, you can directly get the `href` property and not need to use `getAttribute`.

```suggestion
    const href = permalinkElement.href
```

src/shared/util/dom.tsx
```diff
@@ -12,3 +12,31 @@ export function insertAfter(el: HTMLElement, referenceNode: Node): void {
...
+    if (!href) {
+        return new Error(
```
ijsnow: Throwing an error is more common in JS

src/libs/github/file_info.ts
```diff
@@ -64,21 +66,36 @@ export const resolveDiffFileInfo = (codeView: HTMLElement): Observable<FileInfo>
...
         )
-    )
+    }
+
+    const commitID = commitIDFromPermalink({
+        selector: '.js-permalink-shortcut',
+        hrefRegex: new RegExp('^/.*?/.*?/blob/([0-9a-f]{40})/'),
+    })
+
+    if (isErrorLike(commitID)) {
```
ijsnow: Throwing the error inside `commitIDFromPermalink` would prevent the need for this, but you could add a check for whether its not a string here to prevent the need for the separate declaration below

```suggestion
    if (isErrorLike(commitID) || typeof commitID !== 'string') {
```

src/libs/gitlab/file_info.ts
```diff
@@ -29,21 +30,33 @@ const ensureRevisionsAreCloned = (files: Observable<FileInfo>): Observable<FileI
...
+        return throwError(
+            new Error(
+                `Unable to determine the file path of the current file because the current URL (window.location ${
+                    window.location
+                }) does not have a file path.`
             )
         )
-    )
+    }
+    const commitID = commitIDFromPermalink({
```
ijsnow: I purposefully kept DOM stuff outside of this file. Could you move this into `scrape`? Why is this needed if we get the rev from `getFilePageInfo` anyways?

src/libs/gitlab/file_info.ts
```diff
@@ -29,21 +30,33 @@ const ensureRevisionsAreCloned = (files: Observable<FileInfo>): Observable<FileI
...
+        selector: '.js-data-file-blob-permalink-url',
+        hrefRegex: new RegExp('^/.*?/.*?/blob/([0-9a-f]{40})/'),
+    })
+    if (isErrorLike(commitID)) {
```
ijsnow: Same as the github file
```suggestion
    if (isErrorLike(commitID) || typeof commitID !== 'string') {
```

src/libs/gitlab/file_info.ts
```diff
@@ -29,21 +30,33 @@ const ensureRevisionsAreCloned = (files: Observable<FileInfo>): Observable<FileI
...
+        return throwError(commitID)
+    }
+    // TypeScript doesn't realize that `commitID` is not an `Error` here.
+    const stringCommitID: any = commitID
```
ijsnow: Same as the github file. 
```suggestion
    const stringCommitID = commitID as string
```

src/libs/github/file_info.ts
```diff
@@ -64,21 +66,36 @@ export const resolveDiffFileInfo = (codeView: HTMLElement): Observable<FileInfo>
...
+        return throwError(commitID)
+    }
+
+    // TypeScript doesn't realize that `commitID` is not an `Error` here.
+    const stringCommitID: any = commitID
+    return of({
+        repoPath,
+        filePath,
+        commitID: stringCommitID,
+        rev: stringCommitID,
```
ijsnow: This will mean that if someone is looking at a file on a branch, the pretty branch url will be blown away for the commit, which is a regression. I'm not sure where the issue is for this but this was an issue I cause/fixed before. This should use the `rev` from `parseURL`.

src/libs/gitlab/file_info.ts
```diff
@@ -29,21 +30,33 @@ const ensureRevisionsAreCloned = (files: Observable<FileInfo>): Observable<FileI
...
+    return of({
+        repoPath,
+        filePath,
+        commitID: stringCommitID,
+        rev: stringCommitID,
```
ijsnow: This will mean that if someone is looking at a file on a branch, the pretty branch url will be blown away for the commit, which is a regression. I cause this on GitHub before and had to fix it from user complaints. This should use the `rev` from `getFilePageInfo`.

src/shared/backend/graphql.tsx
```diff
@@ -113,61 +132,10 @@ function shouldResponseTriggerRetryOrError(response: any): boolean {
...
- */
-export function queryGraphQLNoRetry(
-    ctx: RequestContext,
-    query: string,
-    variables: any = {},
-    url: string = sourcegraphUrl
-): Observable<QueryResult<IQuery>> {
-    return requestGraphQL(ctx, query, variables, url, false) as Observable<QueryResult<IQuery>>
-}
+export const queryGraphQL = (args: GraphQLRequestArgs) => requestGraphQL(args) as Observable<QueryResult<IQuery>>
```
ijsnow: Can we use generics to not have to cast with `as` in so many places?

src/libs/gitlab/file_info.ts
chrismwendt: Moved to `scrape`. `getFilePageInfo` doesn't resolve to a commit ID, does it? It looks like it only gives you the revision (i.e. it could be a ref like `master` rather than a commit ID).

chrismwendt:
Thanks, @ijsnow, I accepted all of your suggestions except the `.href` one :sparkles:

> The changes look nice but unnecessary for the changes intended in this PR. Could we break that out separately so we can easily merge our PRs and then get those changes in afterwards?

Mind if I try resolving the conflicts? No conflict is too difficult to surmount with the almighty [`diffme`](https://gist.github.com/chrismwendt/d2fed14dbb2c003bfb823005d9d3f8fe) 🔨

chrismwendt:
I resolved the conflicts in `src/shared/backend/graphql.tsx` and `src/shared/backend/server.ts` (there didn't appear to be any fundamental conflicts, just some code shuffling).

src/shared/util/dom.tsx
ijsnow: I would prefer `.href` becuase [I use it elsewhere](https://sourcegraph.com/github.com/sourcegraph/browser-extensions/-/blob/src/libs/gitlab/scrape.ts#L160) but its not worth spending much more time on. I'm fine with `getAtrribute`.

PR #251, opened 2018-10-17
fix: only show one code intel status indicator

chrismwendt:
This hides the legacy status indicator when Sourcegraph extensions are being used.


felixfbecker:
Does this also remove the third one? 

image

src/shared/components/CodeViewToolbar.tsx
```diff
@@ -10,6 +10,7 @@ import {
 import * as React from 'react'
 import { Subscription } from 'rxjs'
 import { ContributableMenu } from 'sourcegraph/module/protocol'
+import storage from '../../browser/storage'
```
ijsnow: I think we should keep usage of browser APIs to a minimum outside of `src/extension` because of the in page phabricator script. Although we allow `observeSync` to just be a noop rather than throw when used in the page, the abstraction in `src/browser` was intended for cross browser support and not being able to run in page and as an extension. 

Is checking `this.props.extensionsController` enough?

src/shared/components/CodeViewToolbar.tsx
```diff
@@ -83,16 +88,18 @@ export class CodeViewToolbar extends React.Component<CodeViewToolbarProps, CodeV
...
-                    key="code-intel-status"
-                    userIsSiteAdmin={currentUser ? currentUser.siteAdmin : false}
-                    repoPath={this.props.repoPath}
-                    commitID={this.props.commitID}
-                    filePath={this.props.filePath}
-                    onChange={this.props.onEnabledChange}
-                    simpleProviderFns={this.props.simpleProviderFns}
-                    site={site}
-                />
+                {!this.state.useExtensions && (
```
ijsnow: As mentioned above, could we just do this instead of using storage in this file?

```suggestion
                {!this.props.extensionsController && (
```

felixfbecker:
Will check once released. Are you planning to cut a release?

PR #267, opened 2018-10-25
fix: access token logic

ijsnow:
The problem: Chrome began rejecting requests from the browser extension options page saying that `authorization` headers weren't allowed from that origin (the chrome extension bundle ID). This PR fixes that by not using tokens on requests coming from the options page or background script. That change brought about another issue which was simply a result of an architecture decision in the PR that introduces access tokens. The problem is that tokens would never be created since they were done on an as-needed basis at the bottom of the funnel for each request. Right before any request went out, we'd check for a token and if it didn't exist we'd create one. Now, we create a token as soon as the user enters their sourcegraph URL and never again.

This commit fixes access token usage by:

- Not setting the `authorization` header when
requesting from the options page.
Chrome suddenly began rejecting the request in
pre-flight checks from the options page.
- _Only_ creating access tokens when the server connection is checked in the
options page.
- Save the access token ID so we can ensure it still exists on the
server.

- [x] Chrome
- [x] Firefox


src/shared/components/options/ConnectionCard.tsx
```diff
@@ -242,6 +248,34 @@ export class ConnectionCard extends React.Component<Props, State> {
...
+                    switchMap(token => fetchCurrentUser(false).pipe(map(user => ({ user, token })))),
+                    filter(propertyIsDefined('user')),
+                    // Get the IDs for all access tokens for the user.
+                    switchMap(({ token, user }) =>
+                        fetchAccessTokenIDs(user.id).pipe(map(usersTokenIDs => ({ usersTokenIDs, user, token })))
+                    ),
+                    // Make sure the token still exists on the server. If it
+                    // does exits, use it, otherwise create a new one.
+                    switchMap(({ user, token, usersTokenIDs }) => {
+                        const tokenExists = !!token && !!usersTokenIDs.find(({ id }) => id === token.id)
```
chrismwendt:
```suggestion
                        const tokenExists = usersTokenIDs.map(({ id }) => id).includes(token.id)
```

src/extension/scripts/background.tsx
```diff
@@ -166,6 +166,19 @@ storage.addSyncMigration((items, set, remove) => {
         set({ accessTokens: {} })
     }
 
+    if (items.accessTokens) {
+        const accessTokens = {}
+
+        for (const key of Object.keys(items.accessTokens)) {
+            const val = items.accessTokens[key]
+            if (typeof val !== 'string' && val.id && val.token) {
+                accessTokens[key] = val
```
chrismwendt: Nitpick: `key`/`val` are pretty vague - it took me a while to realize that `key` is the Sourcegraph URL and `val` is the token.

PR #272, opened 2018-10-25
feat: always use Sourcegraph extensions, remove feature flag

sqs:
Removes the feature flag to toggle between whether to use Sourcegraph extensions and not. Sourcegraph extensions will always be used.

@ijsnow 	I opened this up per https://sourcegraph.slack.com/archives/C07KZF47K/p1540456636000100?thread_ts=1540453431.000100&cid=C07KZF47K. What else is necessary before we can merge it? Are there any other old code paths to remove?

Feel free to take over this PR.


ijsnow:
Rethinking, we should consider waiting until I get at least the second bullet point done here https://github.com/sourcegraph/browser-extensions/issues/273. 

It looks like e2e tests passed in CI but I've ran them 4 times locally now and they've failed 1/4 times in different places. Could be due to wake up time for the lang server extensions?

chrismwendt:

```
cd browser-extensions

cd ../sourcegraph
# and fixup merge conflicts
```

PR #275, opened 2018-10-25
fix: do not display code intelligence hovers on Sourcegraph Extension…

slimsag:
…line decorations

Prior to this change, hovering over text/links provided by a Sourcegraph Extension would trigger the display of the hover 'loading' indicator. On most files, it goes away quickly (just a flicker), but on some it displays an error:

image

After this change, we no longer attempt to display hovers etc. on the DOM elements which Sourcegraph Extensions provide.

Fixes sourcegraph/sourcegraph-extension-api#67

Testing plan:

- Find a line where a Sourcegraph Extension (e.g. codecov or code discussions) provides after-line decorations, hover over it.
- Hover over plain code and confirm code intel hovers also work.

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


src/libs/code_intelligence/code_intelligence.tsx
```diff
@@ -365,6 +365,19 @@ function handleCodeHost(codeHost: CodeHost): Subscription {
...
+                    const originalDOM = dom
+                    dom = {
+                        ...dom,
+                        // If any parent element has the sourcegraph-extension-element
+                        // class then that element does not have any code. We
+                        // must check for "any parent element" because extensions
+                        // create their DOM changes before the blob is tokenized
+                        // into multiple elements.
+                        getCodeElementFromTarget: (target: HTMLElement): HTMLElement | null =>
+                            target.closest('.sourcegraph-extension-element') !== null
```
chrismwendt: I'd factor this out into a variable to enable find refs.

slimsag: Please clarify, I don't understand. What would be moved to a variable? How would it enable "find refs"?

chrismwendt:
```suggestion
export const sourcegraphExtensionElementClass = '.sourcegraph-extension-element'
...
                            target.closest('.' + sourcegraphExtensionElementClass) !== null
```

PR #276, opened 2018-10-26
fix: various Sourcegraph extension line attachment bugs

slimsag:
Please see individual commit messages for details.

Compare changes here with what we do in the Sourcegraph codebase: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@4ac84723437a325405bf073a99f45fae1be9ec43/-/blob/src/repo/blob/LineDecorationAttachment.tsx#L40-62

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


src/libs/code_intelligence/extensions.tsx
```diff
@@ -176,6 +176,14 @@ export const applyDecoration = (
             const link = document.createElement('a')
             link.className = 'sourcegraph-extension-element'
             link.setAttribute('href', url)
+
+            // External URLs should open in a new tab, whereas relative URLs
+            // should not.
+            link.setAttribute('target', /^https?:\/\//.test(url) ? '_blank' : '')
```
chrismwendt: I prefer website to give me the option to open the link in the same tab.

slimsag: I'm going to merge this as-is because I am just reflecting the current state of what we do in Sourcegraph. But I think you should create an issue for this and make an argument for opening links in the same tab. It's not my preference, but we should probably gather more feedback here.

PR #278, opened 2018-10-28
fix: sync z-index for github tree and hover overlay

lukeautry:
This PR changes:

Addresses an issue surfaced in https://github.com/sourcegraph/sourcegraph/issues/242 where a hover overlay gets covered up by the Github tree sidebar. This is a simple stack level bug.

Normally I would say to just update z-index for `.hover-overlay` but it would be really easy to reintroduce this bug with a simple edit to z-index on `.tree-mount`. I extracted out the value to a variables.scss file so that the z-index value be shared. I suspect that variables.scss is not a great place to put this and maybe there's an existing pattern to follow here when values need to be shared between scss files, but I couldn't find anything like that in this project.

Testing plan:

- Shrink browser window so a tooltip is likely to overflow to the left
- Hover over some symbol with information; should be easy to reproduce under the old code
- Verify that overlay now goes over GitHub tree sidebar

I have tested on:

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle

### DIFF
--- a/src/shared/tree/tree.scss
+++ b/src/shared/tree/tree.scss
@@ -1,8 +1,10 @@
+@import '../global-styles/variables.scss';
+
 .tree-mount {
     position: fixed;
     top: 0;
     left: 0;
-    z-index: 2000;
+    z-index: $default-z-index;
     width: 100%;
     pointer-events: none;
     height: 100%;
--- a/src/shared/tree/tree.scss
+++ b/src/shared/tree/tree.scss
@@ -4,7 +4,7 @@
     position: fixed;
     top: 0;
     left: 0;
-    z-index: $default-z-index;
+    z-index: $default-z-index - 1;
     width: 100%;
     pointer-events: none;
     height: 100%;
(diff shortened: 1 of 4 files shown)


src/shared/tree/tree.scss
```diff
@@ -1,8 +1,10 @@
+@import '../global-styles/variables.scss';
+
 .tree-mount {
     position: fixed;
     top: 0;
     left: 0;
-    z-index: 2000;
+    z-index: $default-z-index;
```
ijsnow: From what I can tell, making the z-index be the same for both the hover overlay and the tree mount only fixes the problem because the hover overlay's mount is appended to the end of `document.body` but the tree mount is added further up the DOM tree? It's important to note that it doesn't actually need to be. It could just be added to the bottom making this fix not really deterministic. 

What about subtracting one for the default for this? Or just making them 2000 and 1999?
```suggestion
    z-index: $default-z-index - 1;
```

lukeautry: This is a good point in that restructuring the order of how this stuff gets injected could negate these changes. Pushing up the change.

ijsnow:
Great, thanks!

PR #285, opened 2018-10-29
fix: upgrade @sourcegraph/extensions-client-common

sqs:
This PR changes:

- Upgrades to the newest version of `@sourcegraph/extensions-client-common`.
- The newest version requires a `location` to be passed through.

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle


sqs:
@ijsnow 	I'm going to merge this because I need to incorporate other changes from extensions-client-common.

This PR is incomplete because it does not create a router at the top of the hierarchy and propagate location updates. The `location` prop value in some components might be stale. But the `location` prop is only used by extensions using https://github.com/sourcegraph/sourcegraph-extension-api/pull/109, which is a brand new API that no extensions actually use yet (except for my word count demo). So it's OK if the `location` prop value is stale.

PR #291, opened 2018-10-31
add back client settings for unauthed users only

sqs:
This PR changes:

See commit messages.

Testing plan:

1. Try extension actions that update settings as an authed user.
2. View options page as an authed user and ensure it says you are authed.
3. Delete the auto-added access token and sign out of Sourcegraph.
4. Try extension actions that update settings as an unauthed user. Confirm that the updates are reflected in the options page.
5. View options page as an unauthed user and ensure it does not say you are authed.

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle

### DIFF
--- a/src/shared/backend/headers.tsx
+++ b/src/shared/backend/headers.tsx
@@ -2,7 +2,7 @@ import { Observable, of } from 'rxjs'
 import { map, switchMap } from 'rxjs/operators'
 
 import { getAccessToken } from '../auth/access_token'
-import { isContent, isInPage, isPhabricator } from '../context'
+import { isInPage, isPhabricator } from '../context'
 import { getExtensionVersionSync, getPlatformName, isSourcegraphDotCom } from '../util/context'
 
 /**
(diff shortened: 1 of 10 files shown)


src/shared/backend/headers.tsx
```diff
@@ -24,7 +24,7 @@ export function getHeaders(
     }
 
     return of(url).pipe(
-        switchMap(url => (isContent && useToken && !isSourcegraphDotCom(url) ? getAccessToken(url) : of(undefined))),
+        switchMap(url => (useToken && !isSourcegraphDotCom(url) ? getAccessToken(url) : of(undefined))),
```
sqs: @ijsnow 	This is the most important change for you to review. See the commit message for why I needed to make this change.

sqs:
> Is the difference between this and what we had a few weeks ago that this only applies to unauthed users?

Yes (which means it also omits the messaging.ts, connectAsPage, etc., components).

src/shared/backend/headers.tsx
sqs: What errors were you seeing?

I have confirmed that the options.html page is sending HTTP requests (eg for `graphql?Configuration`) with the Authorization header set correctly. This authenticates the request (confirmed by clearing cookies so it was the only way the request could be authed, and it still works).

PR #314, opened 2018-11-04
refactor: use non-deprecated GraphQL Repository.name (not uri)

sqs:
The Repository.uri field in GraphQL (and in general) is deprecated in favor of Repository.name. This is just a different name for the same value (`name` is a better name than `uri` because the value is not actually a URI, so `uri` is a misnomer). The value that both fields contain is the string of the form `github.com/foo/bar` for a repository.

The GraphQL `Query.repository` resolver also has 2 equivalent args, `name` and `uri`, and this also changes calls to use the former (for the same reason).

There is no change in behavior.

Testing plan:

- Load a GitHub page and ensure the GraphQL queries calling `Query.repository` use the `name` arg name and succeed.

I have tested on:

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Phabricator Bundle

### DIFF
--- a/src/shared/backend/search.tsx
+++ b/src/shared/backend/search.tsx
@@ -79,7 +79,7 @@ export function createSuggestion(item: GQL.SearchSuggestion): Suggestion | null 
         case 'Repository': {
             return {
                 type: 'repo',
-                title: item.uri,
+                title: item.name,
                 url: `/${item.name}`,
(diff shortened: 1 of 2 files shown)


chrismwendt:

```
cd browser-extensions

cd ../sourcegraph
# and fixup merge conflicts
```

